PR #31, opened 2020-03-04
Feature/storybook split & ReadMe relocation

emclaug2:
- POC for splitting stories into smaller pieces for easier Story src navigation.
- Add README section to each story & remove API section
- Bucket Drawer & DrawerLayout
- Bucket Hero & Hero Banner


emclaug2:
Don't merge this yet; this is just a POC to see if the new format seems reasonable.

emclaug2:
Motivation for splitting the code comes from a few places:
1.  This new format is a best-practice introduced in Storybook 5.2+; it's called Component Story Format and has a few benefits, including to add tests to individual states if we ever go down this route. 
2. It's easier to find the source code for a story that's displayed in the Canvas.

px-ptera:
I updated the components and theme dependencies, and stopped changing the 'Notes' section title to 'README'. In the process of getting some feedback on the layout.

jeffvg:
Are we able to fix sticky scroll on the read me's? Currently if you scroll to bottom of any read me file then select a different one you land in the middle or bottom of  the other read me

emclaug2:
> Are we able to fix sticky scroll on the read me's? Currently if you scroll to bottom of any read me file then select a different one you land in the middle or bottom of the other read me

I messed around for about 30 minutes and I can't figure out a way to add this scroll top without another layer of duct tape.  Basically we'd have to  call a scrollTop function after clicking the notes tab & post-content render. 

The container that needs scrolled is this class: 'css-10ro1m'

Few things here I'm not a fan of:
1. Using seemingly-generated class names to perform an action.
2. Post-content rendering actions; no easy way to know when the content drawn.

This is a side effect of rendering the README story as the first thing a user sees; if the user views a Story then views the Notes, the scroll distance is reset.

demos/storybook/src/utils.tsx
```diff
@@ -1,31 +1,92 @@
...
+    },
+    notes: {},
+};
+
+export const getReadMe = (name: string): any => {
+    const md = require(`./../../../docs/${name}`);
+
+    // Locate all relative links that use href syntax and replace them with absolute URLs.
+    md.default = (md.default).replace(/\(.\/.*md\)/g, (substring: string) => {
+        const root = 'https://pxblue-components.github.io/react-dev/';
```
joebochill: Can we update this so it uses the 'root' of the current URL rather than hardcoding react-dev?
i.e.,
- on pxblue-components.github.io/react...it should keep us on the master deployment and not direct us to dev when we click a link
- on localhost, we should stay on localhost when clicking a link, instead of going to react-dev

emclaug2: Updated this, works localy; seems like it'll work in dev

PR #33, opened 2020-03-06
Bug/proptype patch

huayunh:
- Fix prop types
- Fix issues with statusColor & icons caused by forced type cast
- Allow statusColor for nested items


huayunh:
Taking that two slider out to avoid breaking the code... besides, I think those two might not be that critical to demonstrate our api anyway...

docs/Drawer.md
```diff
@@ -133,7 +133,7 @@ The `items` prop of the `DrawerNavGroup` takes a list of items with the followin
 | title           | The text to show on the first line                                                                             | `string`           | yes      |                                  |    
 
 #### NestedNavItem Object
-The `items` property of the NavItem can be nested to create a tree structure with expand/collapse panels. Nested items take the same properties as `NavItem` with the exception of `icon` (nested items cannot use icons).
+The `items` property of the NavItem can be nested to create a tree structure with expand/collapse panels. Nested items take the same properties as `NavItem` with the exception of `icon` (nested items cannot use icons). Besides, the default `expandIcon` prop is `<ArrowDropUp />`.
```
joebochill: We don't need to call out the default icon here...they'll see it when they use the component. Or if you really want to tell them what it is, you can call it out in the API row above.

components/src/core/Drawer/DrawerNavGroup.tsx
```diff
@@ -407,25 +408,37 @@ export const DrawerNavGroup: React.FC<DrawerNavGroupProps> = (props) => {
...
 DrawerNavGroup.propTypes = {
     activeBackgroundColor: PropTypes.string,
     activeFontColor: PropTypes.string,
     activeIconColor: PropTypes.string,
+    activeItem: PropTypes.string,
+    activeBackgroundShape: PropTypes.oneOf(['round', 'square']),
     backgroundColor: PropTypes.string,
     chevron: PropTypes.bool,
-    titleContent: PropTypes.element,
+    titleContent: PropTypes.node,
```
joebochill: Does this work correctly if you pass in a node instead of an element?

huayunh: titleContent has been a ReactNode type when it's declared in type DrawerNavGroupProps. But I tried in the showcase to pass in a `<></>`, and it works, so I think this change is legit.

PR #43, opened 2020-03-12
Feature/list item tag stories refactor

emclaug2:
- Remove doc examples column from API
- Add context story
- Rename stories
- Refactor InfoListTag to not duplicate spread props


demos/storybook/stories/list-item-tag/_module.stories.tsx
```diff
@@ -11,7 +11,9 @@ const listItemTagModule = {
 
 /* Display order goes here */
 export { getReadMeStory } from '../../src/utils';
-export * from './with-diff-colors';
-export * from './with-typography-props';
+export * from './with-basic-usage';
+export * from './with-custom-colors';
+export * from './within-InfoListItem';
```
joebochill: Are we calling them 'within X' or ' with context' ?
Do we want the context stories to come last. Showing it before with full config feels like it breaks the flow of stories building on each other.

emclaug2: I think 'within X' is a little more descriptive in this case than 'with context'. 

If I'm told 'with context', I don't immediately know what the context is.

'within an InfoListItem' tells me that this component should be used with another PX Blue component, instead of me looking through the source code to see what the container component is.  

I'll move the 'context story' to the last spot per story group.

PR #44, opened 2020-03-12
Feature/info list item stories refactor

emclaug2:
- Add a full config story
- Modify story names, knob names, increase story progression
- Update docs to remove examples column


jeffvg:
should we include onClick and ripple for "within a full list" example. 💡

emclaug2:
I left the onClick with ripples off for the "within a full list" example because I didn't interpret it as a clickable list.  In my head I'm thinking a list only appears clickable if there's a chevron on it.

demos/storybook/stories/info-list-item/with-basic-usage.tsx
```diff
@@ -0,0 +1,8 @@
+import { InfoListItem } from '@pxblue/react-components';
+import { text } from '@storybook/addon-knobs';
+import { StoryFnReactReturnType } from '@storybook/react/dist/client/preview/types';
+import React from 'react';
+
+export const withBasicUsage = (): StoryFnReactReturnType => <InfoListItem title={text('title', 'Info List Item')} />;
+
+withBasicUsage.story = { name: 'with basic usage' };
```
joebochill: Are we making this a constant ?

demos/storybook/stories/info-list-item/with-full-config.tsx
```diff
@@ -0,0 +1,34 @@
...
+            dense={boolean('dense', false)}
+            divider={appliedDivider}
+            hidePadding={boolean('hidePadding', false)}
+            ripple={boolean('ripple', true)}
+            onClick={action('clicked')}
+        />
+    );
+};
+
+withFullConfig.story = { name: 'with full config' };
```
joebochill: constant?

PR #53, opened 2020-03-16
Add class overrides to Hero component

emclaug2:
- Add class overrides to Hero
- Classes: [root, icon, values, label]

### DIFF
--- a/components/src/core/Hero/Hero.tsx
+++ b/components/src/core/Hero/Hero.tsx
@@ -4,10 +4,11 @@ import Typography from '@material-ui/core/Typography';
 import * as Colors from '@pxblue/colors';
 import { ChannelValue } from '../ChannelValue';
 import PropTypes from 'prop-types';
+import clsx from 'clsx';
 
 const useStyles = makeStyles((theme: Theme) =>
     createStyles({
-        wrapper: {
+        root: {
             fontSize: '1rem',
             display: 'flex',
             flexDirection: 'column',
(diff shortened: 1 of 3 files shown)


components/src/core/Hero/Hero.tsx
```diff
@@ -93,11 +102,14 @@ export const Hero = (props: HeroProps): JSX.Element => {
             >
                 {icon}
             </span>
-            <span className={classes.values} style={{ fontSize: normalizeFontSize(fontSize) }}>
+            <span
+                className={clsx(defaultClasses.values, classes.values)}
+                style={{ fontSize: normalizeFontSize(fontSize) }}
+            >
                 {!props.children && value && <ChannelValue value={value} units={units} icon={valueIcon} />}
```
joebochill: should we add a class here too?

emclaug2: > Do we have any classes for HeroBanner?

I was going to make a separate PR for HeroBanner but I'll include it in this one.

emclaug2: Are you talking about the span container?  Not sure where to add another class.

joebochill: on the ChannelValue component...should we allow them to pass in a class (i.e., root) that goes on that component (since it now accepts classes)? Or maybe it's a nested object where you can set any of the classes that are exposed from the ChannelValue? Does MUI do anything like this?

emclaug2: We've done something like before in the UserMenu, when we allowed the user the pass in MenuProps.  We could do the same thing here by allowing a ChannelValueProps to be passed in.

joebochill: Or maybe we just leave it alone...they can pass in their own ChannelValue component as a child and style it that way if they really want to customize it

emclaug2: Only thing is that I wouldn't want to expose all the other Props that come along with ChannelValue, maybe we should start exporting the Classes type we've defined in the components so we can use them in places.  Here would be a good opportunity to add a ChannelValueClasses prop to Hero.

PR #55, opened 2020-03-16
Feature/channel value classes

emclaug2:
- Add class overrides to ChannelValue Component
- classes: [root, icon, value, units]

### DIFF
--- a/components/src/core/ChannelValue/ChannelValue.tsx
+++ b/components/src/core/ChannelValue/ChannelValue.tsx
@@ -28,7 +28,15 @@ const styles = makeStyles({
     },
 });
 
+
+type ChannelValueClasses = {
+    root?: string;
+    icon?: string;
+    units?: string;
+    value?: string;
+}
 export type ChannelValueProps = {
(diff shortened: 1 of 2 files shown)


components/src/core/ChannelValue/ChannelValue.tsx
```diff
@@ -65,17 +73,17 @@ export const ChannelValue: React.FC<ChannelValueProps> = (props) => {
...
+                <span className={clsx(defaultClasses.icon, classes.icon)} data-test={'icon'}>
                     {changeIconDisplay(icon)}
                 </span>
             )}
             {prefix && getUnitElement()}
             <Typography
                 variant={'h6'}
                 color={'inherit'}
-                className={combine([classes.text, classes.value])}
+                className={clsx(combine([defaultClasses.text, defaultClasses.value]), classes.value)}
```
joebochill: I think with clsx here, you can drop the combine function

emclaug2: I dropped the use of combine, seems to look the same.

PR #56, opened 2020-03-16
Feature/infolistitem classes

emclaug2:
- Add class overrides to InfoListItem component
- classes: [root, title, subtitle, rightComponent]

- Experimented with spreading props to List components, but removed it.

### DIFF
--- a/components/src/core/InfoListItem/InfoListItem.tsx
+++ b/components/src/core/InfoListItem/InfoListItem.tsx
@@ -20,7 +20,7 @@ import {
     Typography,
     ListItemSecondaryAction,
 } from '@material-ui/core';
-import clsx from "clsx";
+import clsx from 'clsx';
 
 const useStyles = makeStyles((theme: Theme) =>
     createStyles({
@@ -58,11 +58,11 @@ const useStyles = makeStyles((theme: Theme) =>
             flex: '0 0 auto',
             marginLeft: 16,
             display: 'flex',
-            alignItems: 'center'
+            alignItems: 'center',
         },
         chevron: {
-            display: 'flex'
-        }
+            display: 'flex',
+        },
     })
 );
 
@@ -70,10 +70,10 @@ const MAX_SUBTITLE_ELEMENTS = 6;
 
 type InfoListItemClasses = {
     root?: string;
-     title?: string;
-     subtitle?: string;
+    title?: string;
+    subtitle?: string;
     rightComponent?: string;
-}
+};
 
 export type DividerType = 'full' | 'partial';
 export type InfoListItemProps = {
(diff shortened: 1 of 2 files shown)


components/src/core/InfoListItem/InfoListItem.tsx
```diff
@@ -53,16 +54,33 @@ const useStyles = makeStyles((theme: Theme) =>
             lineHeight: 1.3,
             color: 'inherit',
         },
+        rightComponent: {
+            flex: '0 0 auto',
+            marginLeft: 16,
```
joebochill: theme.spacing(2)

components/src/core/InfoListItem/InfoListItem.tsx
```diff
@@ -144,22 +163,18 @@ export const InfoListItem: React.FC<InfoListItemProps> = (props) => {
 
     const getRightComponent = (): JSX.Element | undefined => {
         if (rightComponent) {
-            return (
-                <div style={{ flex: '0 0 auto', marginLeft: 16, display: 'flex', alignItems: 'center' }}>
-                    {rightComponent}
-                </div>
-            );
+            return <div className={clsx(defaultClasses.rightComponent, classes.rightComponent)}>{rightComponent}</div>;
```
joebochill: do we want to use the same class (rightComponent) for chevron and rightComponent since chevron is just a shortcut for rightcomponent?

components/src/core/InfoListItem/InfoListItem.tsx
```diff
@@ -144,22 +163,18 @@ export const InfoListItem: React.FC<InfoListItemProps> = (props) => {
...
+                <ListItemSecondaryAction className={defaultClasses.chevron}>
                     <Chevron color={'inherit'} />
                 </ListItemSecondaryAction>
             );
         }
     };
 
     const interpunct = (): JSX.Element => (
-        <Typography className={`${classes.withSmallMargins} ${classes.separator}`}>
+        <Typography className={`${defaultClasses.withSmallMargins} ${defaultClasses.separator}`}>
```
joebochill: we can use clsx here

joebochill: do we want to expose a class here?

components/src/core/InfoListItem/InfoListItem.tsx
emclaug2: Updated this to use theme

components/src/core/InfoListItem/InfoListItem.tsx
emclaug2: I've removed the withSmallMargins class, combined it with the separator class, and made a class override for it.

components/src/core/InfoListItem/InfoListItem.tsx
emclaug2: rightComponent has different styles applied to it than chevron does; I think it would add unnecessary margins if we applied the rightComponent class.

PR #52, opened 2020-03-16
Drawer Inheritable Props

huayunh:
Fixes #28, fixes #34  .

### Lift/extend 18 props

Apply `PXBlueDrawerInheritableProperties` and `PXBlueDrawerInheritableGroupProperties` to `Drawer`, `DrawerBody`, `DrawerNavGroupProps` and (Nested)`NavItem`:

| Name                      | Description                                                  | Type               | Default                                                      |
| ------------------------- | ------------------------------------------------------------ | ------------------ | ------------------------------------------------------------ |
| activeItemBackgroundColor | Background color for the 'active' item                       | `string`           | (theme.palette.type === 'light' ? primary50Color : theme.palette.primary.main) |
| activeItemBackgroundShape | shape of the active item background                          | `'round'|'square'` | round                                                        |
| activeItemFontColor       | Font color for the 'active' item                             | `string`           | (theme.palette.type === 'light' ? theme.palette.primary.main : theme.palette.primary.contrastText) |
| activeItemIconColor       | Icon color for the 'active' item                             | `string`           | (theme.palette.type === 'light' ? theme.palette.primary.main : theme.palette.primary.contrastText) |
| chevron                   | Whether to have chevrons for all menu items                  | `boolean`          |                                                              |
| collapseIcon              | Icon used to collapse drawer<br />default is expandIcon rotated 180 degrees | `JSX.Element`      | `expandIcon` rotated 180 degree                              |
| divider                   | Whether to show a line between all items                     | `boolean`          | true                                                         |
| expandIcon                | Icon used to expand drawer                                   | `JSX.Element`      | `<ExpandLess />` at top-level, `<ArrowDropUp />` otherwise   |
| hidePadding               | Whether to hide the paddings reserved for menu item icons    | `boolean`          |                                                              |
| itemFontColor             | The color used for the item text                             | `string`           | gray[500]                                                    |
| itemIconColor             | The color used for the icon                                  | `string`           | gray[500]                                                    |
| onItemSelect              | internal API<br />will apply to all menu items when onClick  | `() => void`       |                                                              |
| ripple                    | Whether to apply material ripple effect to items             | `boolean`          | true                                                         |

And there are some properties that applies to only `Drawer`, `DrawerBody`, and `DrawerNavGroupProps` (so not `NavItem`):

| Name                  | Description                                                  | Type      | Default                                                      |
| --------------------- | ------------------------------------------------------------ | --------- | ------------------------------------------------------------ |
[...]

PR #58, opened 2020-03-17
Feature/drawer classes

emclaug2:
- Add class overrides to Drawer


emclaug2:
InfoListItem Classes PR needs to be merged before this can be completed; DrawerNavGroups needs to be able to pass class overrides.

components/src/core/Drawer/DrawerNavGroup.tsx
```diff
@@ -273,7 +278,7 @@ function NavigationListItem(
                         expandHandler();
                     }
                 }}
-                className={`${classes.expandIcon} ${!collapseIcon && expanded ? classes.expanded : ''}`}
+                className={`${defaultClasses.expandIcon} ${!collapseIcon && expanded ? defaultClasses.expanded : ''}`}
```
joebochill: shall we allow customization here?

components/src/core/Drawer/DrawerNavGroup.tsx
```diff
@@ -291,10 +296,19 @@ function NavigationListItem(
...
+                classes.listItemContainer,
+                active && defaultClasses.listItemNoHover
+            )}
+        >
             {active && (
                 <div
-                    className={`${classes.active} ${activeBackgroundShape === 'square' ? classes.square : ''}`}
+                    className={`${defaultClasses.active} ${
+                        activeBackgroundShape === 'square' ? defaultClasses.square : ''
+                    }`}
```
joebochill: you can use clsx to conditionally apply a class.
Also, do we want to allow customization here?

joebochill:
I want to merge #52 in prior to this (if it's ready), since it'll probably be easier for you to handle merge conflicts than Huayun

components/src/core/Drawer/DrawerNavGroup.tsx
emclaug2: I added a customization to the active class, I left the square class uncustomizable because they can opt to not sure it.

components/src/core/Drawer/DrawerNavGroup.tsx
emclaug2: Added expandIcon class

components/src/core/Drawer/DrawerNavGroup.tsx
emclaug2: Used clsx to make the class concatentation look a lot prettier

emclaug2:
> I want to merge #52 in prior to this (if it's ready), since it'll probably be easier for you to handle merge conflicts than Huayun

Sounds good to me!  Work is wrapped up for this PR, until the merging happens.

PR #59, opened 2020-03-18
Feature/fix drawerlayout

emclaug2:
- Navigate to the top of page when user selects an item from the Drawer
- Fix scroll issue on mobile, the URL top bar would never disappear
- DrawerLayout API remains the same, no action required from users.


components/src/core/DrawerLayout/DrawerLayout.tsx
```diff
@@ -1,36 +1,43 @@
...
+import { makeStyles, Theme, useTheme } from '@material-ui/core/styles';
 import PropTypes from 'prop-types';
 
 export type DrawerLayoutProps = {
     children: React.ReactNode;
     drawer: ReactNode;
 };
 
-const useStyles = makeStyles({
+const useStyles = makeStyles((theme: Theme) => ({
```
joebochill: can we do this in the same fashion we've done elsewhere:
```
const useStyles = makeStyles((theme: Theme) =>
    createStyles({
```

components/src/core/Drawer/Drawer.tsx
```diff
@@ -238,6 +249,7 @@ export const DrawerComponent: React.FC<DrawerComponentProps> = (props) => {
                     <div
                         className={clsx(defaultClasses.smooth, classes.smooth)}
                         style={{ width: containerWidth, height: '100%' }}
+                        id={'pxblue-drawer-content-width'}
```
joebochill: shall we prefix/namespace these fixed ids in some way to ensure they don't get stomped on (unlikely, but I'm just wondering), e.g.,:
```
id={'@@pxblue/drawer-content-width'}
```

components/src/core/Drawer/Drawer.tsx
```diff
@@ -182,6 +182,7 @@ export const DrawerComponent: React.FC<DrawerComponentProps> = (props) => {
                     titleColor,
                     drawerOpen: isDrawerOpen(),
                     onItemSelect: () => {
+                        window.scrollTo(0, 0);
```
joebochill: what if the user's onClick event is not a navigation function to a new route?

docs/Drawer.md
```diff
@@ -288,3 +288,11 @@ import { Drawer, DrawerLayout } from '@pxblue/react-components';
...
 </DrawerLayout>
 ```
+
+
+## Limitations
+The PX Blue react component library is currently is not dependent on a single Router provider.  
+Because of this limitation, there are a few things additional steps that need happen for users to have an optimal routing experience.
+It is recommended that when new page content loads, the window is adjusted so the top of the page is shown.
+If your application supports routing that occurs outside the scope of the Drawer, the window will have to be manually adjusted.
+If you are using React-Router, the [documentation](https://reacttraining.com/react-router/web/guides/scroll-restoration) references this situation and has a solution.
```
joebochill:
```suggestion
> **Note on Scrolling**: When using client-side routing in your application, you may notice that the window scroll position does not reset when navigating to new routes. To address this issue, you will need to manually update the scroll position when new pages are loaded. If you are using React Router they have [several examples](https://reacttraining.com/react-router/web/guides/scroll-restoration) on how to implement this in your application.
```

emclaug2: Updated the docs to use this explanation

components/src/core/Drawer/Drawer.tsx
emclaug2: I introduced a new inheritable prop that allows the user to control if they want the window to scroll top on navigation.  The default is yes.

components/src/core/Drawer/Drawer.tsx
emclaug2: Good idea, I updated the ID used.

components/src/core/DrawerLayout/DrawerLayout.tsx
emclaug2: Yep I've updated this.

emclaug2:
I want #61 to be merged in before this PR.  I can handle the merging and there might be some tweaks required afterwards.

emclaug2:
This PR also has a defect fix for the InfoListItem component.   The default background color of the InfoListItem was accidentally changed from 'inherit' to use a color from the applied theme palette.

emclaug2:
I got the Showcase demo to work as expected on mobile.  When I integrate this component into the new docit, there's some funky behavior on mobile viewports.  I suspect this is because docit might not be integrating it correctly.

joebochill:
> I got the Showcase demo to work as expected on mobile. When I integrate this component into the new docit, there's some funky behavior on mobile viewports. I suspect this is because docit might not be integrating it correctly.

Doc-it has some hacks in place to work with the old version...it'll need to be updated

components/package.json
```diff
@@ -43,9 +43,8 @@
   },
   "peerDependencies": {
     "@material-ui/core": "^4.7.2",
-    "react": "^16.8.6",
-    "react-dom": "^16.8.6",
-    "@pxblue/colors": "^2.0.0"
+    "@pxblue/colors": "^2.0.0",
+    "react": "^16.8.6"
```
joebochill: do we not have a peer dependency on react-dom anymore?

components/src/core/Drawer/Drawer.tsx
```diff
@@ -215,6 +218,15 @@ export const DrawerComponent: React.FC<DrawerComponentProps> = (props) => {
...
     const contentWidth = width || defaultContentWidth;
 
+    // TODO: Find a better way to rerender.
+   // Initial draw doesn't have parent container rendered so margin not applied on first load.
+    setTimeout(() => {
+      const content = document.getElementById('@@pxb-drawerlayout-content');
+      if (content) {
+         content.style.marginLeft = `${containerWidth}px`;
+      }
+    });
```
joebochill: - might this be better as a useEffect hook to only run when the open property changes?
- would this logic make more sense in the DrawerLayout...presumably, they may be using the Drawer component without one

components/src/core/Drawer/DrawerNavGroup.tsx
```diff
@@ -195,6 +196,10 @@ function NavigationListItem(
 
     // row action
     const action = (): void => {
+        // Don't scroll to top of page when clicking expandable items.
+        if (!disableScrollTop && !items) {
+            window.scrollTo(0, 0);
+        }
```
joebochill: Expandable drawer items can still have a navigation action (there are separate events for clicking the item and clicking the expand icon)

demos/showcase/src/App.js
```diff
@@ -216,7 +216,7 @@ export default () => {
             }
         >
             <div style={{ height: '100%', display: 'flex', flexDirection: 'column' }}>
-                <AppBar position={'static'} color={'primary'}>
+                <AppBar position={'sticky'} color={'primary'}>
```
joebochill: Can you add routing to the showcase so we can see the effects of the scroll behavior?

docs/Drawer.md
```diff
@@ -266,6 +266,7 @@ The following props can be set at any level in the drawer hierarchy (`Drawer`, `
 | chevron                   | Whether to have chevrons for all menu items                | `boolean`          | no       |                                                            |
 | collapseIcon              | Icon used to collapse drawer                               | `JSX.Element`      | no       | `expandIcon` rotated 180 degrees                           |
 | divider                   | Whether to show a line between all items                   | `boolean`          | no       | true                                                       |
+| disableScrollTop          | Whether to scroll to top of the page on navigation         | `boolean`          | no       |                                                            |
```
joebochill:
```suggestion
| disableScrollTop          | Disable automatic scroll reset when clicking a NavItem         | `boolean`          | no       |                                                            |
```

components/package.json
emclaug2: Great catch! This was an accidental edit.

components/src/core/Drawer/Drawer.tsx
emclaug2: I agree it makes mores sense to move this logic to the DrawerLayout component, but it makes the complexity a bit harder.  There's one situation where I'm having trouble getting the refresh logic to work.  Whenever the Drawer NavItems are hovered, the Drawer expands.  When this happens, the DrawerLayout now needs to know that it a redraw is required.  Not sure on how to do this functionality.

emclaug2: I kept the redraw logic in the Drawer and used useEffect.  This works.

emclaug2: The attempted redrawer logic refactor from before was put in a branch called feature/fix-drawerlout-relocated-redraw

components/src/core/Drawer/DrawerNavGroup.tsx
emclaug2: I removed this comment and moved the window.scrollTop logic to be grouped with the onClick action.

demos/showcase/src/App.js
emclaug2: Yep added this to the demo.

components/src/core/Drawer/Drawer.tsx
```diff
@@ -215,6 +215,13 @@ export const DrawerComponent: React.FC<DrawerComponentProps> = (props) => {
     const containerWidth = isDrawerOpen() ? width || defaultContentWidth : defaultNavigationRailWidth;
     const contentWidth = width || defaultContentWidth;
 
+    useEffect(() => {
+        const content = document.getElementById('@@pxb-drawerlayout-content');
+        if (content) {
+            content.style.marginLeft = `${containerWidth}px`;
+        }
+    });
```
joebochill:
```suggestion
    useEffect(() => {
        const content = document.getElementById('@@pxb-drawerlayout-content');
        if (content) {
            content.style.marginLeft = `${containerWidth}px`;
        }
    },[containerWidth]);
```

demos/showcase/src/router/index.js
```diff
@@ -0,0 +1,33 @@
...
+import { DrawerLayout } from '@pxblue/react-components';
+import { NavigationDrawer } from './NavigationDrawer';
+import { App } from '../App';
+import { SharedAppBar } from '../components/SharedAppBar';
+
+export const MainRouter = () => {
+    const [open, setOpen] = useState(true);
+    return (
+        <Router>
+            <DrawerLayout drawer={<NavigationDrawer open={open} setOpen={setOpen} />}>
```
joebochill:
```suggestion
const ScrollToTop = () => {
    const { pathname } = useLocation();

    useEffect(() => {
        window.scrollTo(0, 0);
    }, [pathname]);

    return null;
}

export const MainRouter = () => {
    const [open, setOpen] = useState(true);
    return (
        <Router>
            <ScrollToTop/>
            <DrawerLayout drawer={<NavigationDrawer open={open} setOpen={setOpen} />}>
```

demos/showcase/src/router/index.js
```diff
@@ -0,0 +1,44 @@
+import React, { useState } from 'react';
```
joebochill:
```suggestion
import React, { useState, useEffect } from 'react';
```

demos/showcase/src/router/index.js
```diff
@@ -0,0 +1,44 @@
...
+import { BrowserRouter as Router, Switch, Route, Redirect } from 'react-router-dom';
+import { DrawerLayout } from '@pxblue/react-components';
+import { NavigationDrawer } from './NavigationDrawer';
+import { App } from '../App';
+import { SharedAppBar } from '../components/SharedAppBar';
+
```
joebochill:
```suggestion
import { useLocation } from "react-router-dom";

```

emclaug2: Added missing import

demos/showcase/src/router/index.js
emclaug2: Added missing import

PR #60, opened 2020-03-18
Update to use themes with white input backgrounds

emclaug2:
- Update to use latest storybook theme, which makes input field bg color white
- Remove style overrides for white input field bg color


emclaug2:
@huayunh Can you doublecheck to see if I didn't remove any classes that weren't targeted for the knob input fields?

huayunh:
You guys are moving fast on this but yes, you removed the right thing. Thanks for doing this.

PR #61, opened 2020-03-19
Add variants to drawer and drawer layout

### LINKED ISSUE #22: Drawer should support multiple variants
The drawer should support the variants:
- `temporary`: Mobile-style, only visible when open, no navigation rail.
- `persistent`(default): Current behavior. Can be expanded, but when collapsed, shows the navigation rail.
- `permanent`: Always visible in the fully extended form.

Currently, we only support the `persistent` behavior described above.

This will be necessary for the doc-site redesign to lock the drawer open

#### Additional Context
Both the `persistent` and `permanent` behaviors above will necessitate using the `variant={'permanent'}` on the underlying MUI Drawer component. The slight difference from MUI will need to be mentioned in the docs.

huayunh:
Fixes #22 .

- Add `variant` prop to `Drawer`
- Add `variantBreakpoint` prop to `DrawerLayout`
- Added an extra story to demonstrate the variant
- Updated test, showcase and change log

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Change Log
 
 ## v3.0.0
-- Adds support for nested items in the Drawer component; adjusted Drawer API.
+- Adds support for nested items in the Drawer component; adjusted Drawer and DrawerLayout API.
 
 **Breaking Changes:** 
 - A few props got renamed to avoid further ambiguities: 
@@ -40,6 +40,7 @@
         ]}
     />
     ```
+- DrawerLayout's `drawer` prop now takes `JSX.Element` instead of `ReactNode`.
 
 ## v2.1.0
 - Adds InfoListTag Component
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Change Log
 
 ## v3.0.0
-- Adds support for nested items in the Drawer component; adjusted Drawer and DrawerLayout API.
+- Adds support for nested items in the Drawer component; 
+- Additional styling props added to Drawer; 
 
 **Breaking Changes:** 
 - A few props got renamed to avoid further ambiguities: 
@@ -40,7 +41,6 @@
         ]}
     />
     ```
-- DrawerLayout's `drawer` prop now takes `JSX.Element` instead of `ReactNode`.
 
 ## v2.1.0
 - Adds InfoListTag Component
--- a/components/src/core/Drawer/Drawer.tsx
+++ b/components/src/core/Drawer/Drawer.tsx
@@ -122,12 +122,14 @@ export const DrawerComponent: React.FC<DrawerComponentProps> = (props) => {
         ripple,
         titleColor,
         width,
-        variant,
         ...drawerProps // for Material-UI's Drawer component
     } = props;
 
+    const variant = props.variant || 'persistent'; // to allow drawerLayout to override this
+
     const isDrawerOpen = (): boolean => {
         if (variant === 'persistent') return hover || open;
+        if (variant === 'permanent') return true;
         return open;
     };
 
--- a/components/src/core/Drawer/Drawer.tsx
+++ b/components/src/core/Drawer/Drawer.tsx
@@ -1,39 +1,27 @@
 import React, { useState } from 'react';
-import { createStyles, makeStyles, Theme, useTheme } from '@material-ui/core/styles';
+import { makeStyles, useTheme } from '@material-ui/core/styles';
 import { Drawer, DrawerProps } from '@material-ui/core';
-import Hidden from '@material-ui/core/Hidden';
 import PropTypes from 'prop-types';
 import { DrawerBodyProps } from './DrawerBody';
 import clsx from 'clsx';
 
-const useStyles = makeStyles((theme: Theme) =>
-    createStyles({
-        paper: {
-            overflow: 'hidden',
-            position: 'unset',
-        },
-        paperMobile: {
-            maxWidth: '85%',
-            width: theme.spacing(45),
-        },
-        smooth: {
-            transition: 'width 175ms cubic-bezier(.4, 0, .2, 1)',
-        },
-        content: {
-            display: 'flex',
-            flexDirection: 'column',
-            height: '100%',
-           width: '100%'
-        },
-    })
-);
+const useStyles = makeStyles({
+    paper: {
+        overflow: 'hidden',
+        position: 'unset',
+    },
+    content: {
+        display: 'flex',
+        flexDirection: 'column',
+        height: '100%',
+        width: '100%',
+    },
+});
 
 type DrawerClasses = {
     root?: string;
     content?: string;
-    drawer?: string;
     paper?: string;
-    smooth?: string;
 };
 
 // type shared by Drawer, DrawerBody, DrawerNavGroup, NestedNavItem
@@ -134,10 +122,14 @@ export const DrawerComponent: React.FC<DrawerComponentProps> = (props) => {
         ripple,
         titleColor,
         width,
+        variant,
         ...drawerProps // for Material-UI's Drawer component
     } = props;
 
-    const isDrawerOpen = (): boolean => hover || open;
+    const isDrawerOpen = (): boolean => {
+        if (variant === 'persistent') return hover || open;
+        return open;
+    };
 
     const findChildByType = (type: string): JSX.Element[] =>
         React.Children.map(props.children, (child: any) => {
@@ -215,44 +207,28 @@ export const DrawerComponent: React.FC<DrawerComponentProps> = (props) => {
         </>
     );
 
-    const getMobileNavigationMenu = (): JSX.Element => (
-        <Drawer {...props} open={isDrawerOpen()} classes={{ paper: clsx(defaultClasses.paperMobile, props.classes.paper) }}>
-            <div className={clsx(defaultClasses.smooth, classes.smooth, defaultClasses.content, classes.content)}>
+    const defaultContentWidth = theme.spacing(45);
+    const defaultNavigationRailWidth = theme.spacing(7);
+
+    const containerWidth = isDrawerOpen() ? width || defaultContentWidth : defaultNavigationRailWidth;
+    const contentWidth = width || defaultContentWidth;
+
+    return (
+        <Drawer
+            {...drawerProps}
+            variant={variant === 'temporary' ? variant : 'permanent'}
+            open={isDrawerOpen()}
+            classes={{ paper: clsx(defaultClasses.paper, props.classes.paper) }}
+            style={{
+                minHeight: '100%',
+                width: containerWidth,
+                transition: 'width 175ms cubic-bezier(.4, 0, .2, 1)',
+            }}
+        >
+            <div className={clsx(defaultClasses.content, classes.content)} style={{ width: contentWidth }}>
                 {getDrawerContents()}
             </div>
         </Drawer>
-    );
-
-    const getDesktopNavigationMenu = (): JSX.Element => {
-        const containerWidth = isDrawerOpen() ? width || theme.spacing(45) : theme.spacing(7);
-        const contentWidth = width || theme.spacing(45);
-        return (
-            <>
-                <Drawer
-                    {...drawerProps}
-                    variant={'permanent'}
-                    open={isDrawerOpen()}
-                    classes={{ paper: clsx(defaultClasses.paper, props.classes.paper) }}
-                    style={{ minHeight: '100%' }}
-                >
-                    <div
-                        className={clsx(defaultClasses.smooth, classes.smooth)}
-                        style={{ width: containerWidth, height: '100%' }}
-                    >
-                        <div className={clsx(defaultClasses.content, classes.content)} style={{ width: contentWidth }}>
-                            {getDrawerContents()}
-                        </div>
-                    </div>
-                </Drawer>
-            </>
-        );
-    };
-
-    return (
-        <>
-            <Hidden smUp>{getMobileNavigationMenu()}</Hidden>
-            <Hidden xsDown>{getDesktopNavigationMenu()}</Hidden>
-        </>
     );
 };
 
@@ -285,15 +261,10 @@ DrawerComponent.propTypes = {
         root: PropTypes.string,
         content: PropTypes.string,
         paper: PropTypes.string,
-        paperMobile: PropTypes.string,
-        smooth: PropTypes.string,
     }),
     open: PropTypes.bool.isRequired,
     width: PropTypes.number,
     ...PXBlueDrawerNavGroupInheritablePropertiesPropTypes,
-};
-DrawerComponent.defaultProps = {
-    variant: 'permanent',
 };
 DrawerComponent.defaultProps = {
     classes: {},
--- a/components/src/core/Drawer/DrawerNavGroup.tsx
+++ b/components/src/core/Drawer/DrawerNavGroup.tsx
@@ -73,8 +73,8 @@ const useStyles = makeStyles((theme: Theme) =>
             transitionDuration: '300ms',
             cursor: 'inherit',
             display: 'flex',
-            height: 48,
-            width: 48,
+            height: theme.spacing(6),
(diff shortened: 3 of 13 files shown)


huayunh:
I can do 
```typescript
    const getDrawerContent = (): JSX.Element[] | JSX.Element => {
        if (drawer.type && drawer.type.displayName === 'PXBlueDrawer') {
            return Object.keys(variantBreakpoints).map(
                (bp): JSX.Element =>
                    useMediaQuery(bp) && <Drawer key={bp} variant={variantBreakpoints[bp]} {...drawer.props} />
                )
        } 
        return drawer;
    }
    return (
        <div className={classes.root}>
            <div className={classes.drawer}>
                { getDrawerContent() }
            </div>
            <div className={classes.content}>{children}</div>
        </div>
    );
```

emclaug2:
> ```ts
> getDrawerContent
> ```

Maybe look to the Drawer component for how it searches the children by DisplayName and clones the element.  I'm not sure if this will preserve the props if the user passes in a FC that returns a Drawer.

components/src/core/Drawer/Drawer.tsx
```diff
@@ -285,16 +261,11 @@ DrawerComponent.propTypes = {
...
-        smooth: PropTypes.string,
     }),
     open: PropTypes.bool.isRequired,
     width: PropTypes.number,
     ...PXBlueDrawerNavGroupInheritablePropertiesPropTypes,
 };
-DrawerComponent.defaultProps = {
-    variant: 'permanent',
-};
 DrawerComponent.defaultProps = {
```
joebochill: default variant = persistent?

components/src/core/Drawer/DrawerNavGroup.tsx
```diff
@@ -362,7 +365,7 @@ export const DrawerNavGroup: React.FC<DrawerNavGroupProps> = (props) => {
     return (
         <>
             <List
-                style={{ backgroundColor }}
+                style={{ backgroundColor, paddingBottom: 0 }}
```
joebochill: padding top 0 too?

components/src/core/DrawerLayout/DrawerLayout.tsx
```diff
@@ -1,10 +1,20 @@
...
+import { DrawerProps, useMediaQuery } from '@material-ui/core';
+import { Drawer } from '../Drawer';
 
 export type DrawerLayoutProps = {
+    // Page's body
     children: React.ReactNode;
-    drawer: ReactNode;
+
+    // Drawer component to be embedded
+    drawer: JSX.Element;
```
joebochill: I'll echo Evan's comment that we need to to be able to do this by directly passing in a drawer, or through a FC

demos/showcase/src/App.js
```diff
@@ -107,14 +109,20 @@ export default () => {
...
                                 },
                                 {
                                     title: titleList[1],
                                     itemID: titleList[1],
                                     subtitle: '2 Alarms',
                                     icon: <NotificationsActive />,
-                                    onClick: () => setRoute(1),
+                                    onClick: () => {
+                                        setRoute(1);
+                                        if (xsDown) setOpen(false);
```
joebochill: do we need to do this for all of them, or should we do it even for the persistent drawer?

docs/Drawer.md
```diff
@@ -35,6 +35,7 @@ import { Drawer, DrawerHeader, DrawerSubheader, DrawerBody, DrawerNavGroup, Draw
 | Prop Name                       | Description                                                | Type      | Required | Default  |          
 |---------------------------------|------------------------------------------------------------|-----------|----------|----------|
 | open                            | Controls the open/closed state of the drawer               | `boolean` | yes      |          |
+| variant                         | The variant to use                                         | `'permanent' | 'persistent' | 'temporary'`| no |
```
joebochill: we should add a blurb/callout explaining the difference between our implementation of 'persistent' vs. what MUI does

docs/Drawer.md
```diff
@@ -278,11 +276,34 @@ The `DrawerLayout` component is used to provide the appropriate resizing behavio
     <img width="100%" style="max-width: 600px" alt="DrawerLayout with labels" src="./images/drawerLayout.png">
 </div>
 
+### DrawerLayout API
```
joebochill: Usage should come before API, per how other docs are laid out

docs/Drawer.md
```diff
@@ -278,11 +276,34 @@ The `DrawerLayout` component is used to provide the appropriate resizing behavio
...
+
+<div style="overflow: auto;">
+
+| Prop Name          | Description                                           | Type              | Required  | Default |
+|--------------------|-------------------------------------------------------|-------------------|-----------|---------|
+| children           | Page's body content                                   | `React.ReactNode` | yes       |         |   
+| drawer             | Drawer component to be embedded                       | `JSX.Element`     | yes       |         |   
+| variantBreakpoints | Change the drawer variant according to the breakpoint | `{[key: string]: DrawerProps['variant']}`| no | See below |
```
joebochill: perfect opportunity for an example column :P

joebochill:
I like how small you've managed to keep this PR :)

demos/storybook/stories/drawer/with-different-variants.tsx
```diff
@@ -0,0 +1,21 @@
+import { Drawer, DrawerBody, DrawerHeader, DrawerNavGroup } from '@pxblue/react-components';
+import { boolean, select } from '@storybook/addon-knobs';
+import { StoryFnReactReturnType } from '@storybook/react/dist/client/preview/types';
+import React from 'react';
+import { DrawerStoryContext } from './util';
+import { navGroupItems1 } from './with-basic-config';
+import { Menu } from '@material-ui/icons';
+
+export const withDifferentVariants = (context: DrawerStoryContext): StoryFnReactReturnType => (
+    <Drawer
```
joebochill: Might be more useful to show this in the context of a DrawerLayout so it has more context re: how it behaves with respect to other content.

joebochill:
> > ```ts
> > getDrawerContent
> > ```
> 
> Maybe look to the Drawer component for how it searches the children by DisplayName and clones the element. I'm not sure if this will preserve the props if the user passes in a FC that returns a Drawer.

Yeah, this is actually pretty tricky, because the user could theoretically pass anything in that they want (and we're trying to override the props of something somewhere down in the tree). We sort of want to limit them either to a `<Drawer/>` or another component whose first element in the return statement is a `<Drawer/>`. Maybe we can see if it's a drawer, and if not, see if the first `child` is a drawer and clone that like evan suggests?

components/src/core/Drawer/Drawer.tsx
```diff
@@ -134,10 +122,14 @@ export const DrawerComponent: React.FC<DrawerComponentProps> = (props) => {
         ripple,
         titleColor,
         width,
+        variant,
         ...drawerProps // for Material-UI's Drawer component
     } = props;
 
-    const isDrawerOpen = (): boolean => hover || open;
+    const isDrawerOpen = (): boolean => {
+        if (variant === 'persistent') return hover || open;
```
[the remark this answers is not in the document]
huayunh: Ah. I thought persistent means the drawer can auto open / collapse on hover.

components/src/core/Drawer/Drawer.tsx
huayunh: This default thing will override the default variant we provided in the drawerLayout, so I'd have to take them out.

components/src/core/Drawer/DrawerNavGroup.tsx
huayunh: I think only paddingBottom had the padding issue.

docs/Drawer.md
huayunh: Will be a huge column, sir. I prefer not.

components/src/core/Drawer/Drawer.tsx
huayunh: Let me do that inside the Drawer component...

emclaug2:
> > > ```ts
> > > getDrawerContent
> > > ```
> > 
> > 
> > Maybe look to the Drawer component for how it searches the children by DisplayName and clones the element. I'm not sure if this will preserve the props if the user passes in a FC that returns a Drawer.
> 
> Yeah, this is actually pretty tricky, because the user could theoretically pass anything in that they want (and we're trying to override the props of something somewhere down in the tree). We sort of want to limit them either to a `<Drawer/>` or another component whose first element in the return statement is a `<Drawer/>`. Maybe we can see if it's a drawer, and if not, see if the first `child` is a drawer and clone that like evan suggests?

@joebochill  Do you know if there's a way that we can update the `drawer` prop to specify it's either a Drawer or FC that returns a Drawer?  I haven't seen anything like this so far.

huayunh:
> Maybe we can see if it's a drawer, and if not, see if the first child is a drawer and clone that like evan suggests?

Does this drawer check have to be limited to a `PXBlueDrawer` or including any thing with a `Drawer` displayName?

demos/showcase/src/App.js
huayunh: I think the persistent drawer is already auto collapsing itself (through the onItemSelect function).

components/src/core/Drawer/Drawer.tsx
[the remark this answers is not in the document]
joebochill: **permanent** = always open
**persistent** = can be opened/closed...nav rail when closed, hover makes it expand temporarily
**temporary** = can be opened/closed...invisible when closed (mobile style when open)

components/src/core/Drawer/DrawerNavGroup.tsx
joebochill: ok just wondering...most places we use List we do padding: 0 (not sure why MUI puts padding on their Lists by default)

docs/Drawer.md
joebochill: only kidding

components/src/core/Drawer/Drawer.tsx
joebochill: > > Maybe we can see if it's a drawer, and if not, see if the first child is a drawer and clone that like evan suggests?
> 
> Does this drawer check have to be limited to a `PXBlueDrawer` or including any thing with a `Drawer` displayName?

Probably a PX Blue drawer only...Our DrawerLayout is not going to work with a MUI drawer (ask @emclaug2 about all the trouble he had with the showcase trying to do that by accident)

joebochill:
> > Maybe we can see if it's a drawer, and if not, see if the first child is a drawer and clone that like evan suggests?
> 
> Does this drawer check have to be limited to a `PXBlueDrawer` or including any thing with a `Drawer` displayName?

Probably a PX Blue drawer only...Our DrawerLayout is not going to work with a MUI drawer (ask @emclaug2 about all the trouble he had with the showcase trying to do that by accident)

docs/Drawer.md
huayunh: r/wooooooosh

joebochill:
> > > > ```ts
> > > > getDrawerContent
> > > > ```
> > > 
> > > 
> > > Maybe look to the Drawer component for how it searches the children by DisplayName and clones the element. I'm not sure if this will preserve the props if the user passes in a FC that returns a Drawer.
> > 
> > 
> > Yeah, this is actually pretty tricky, because the user could theoretically pass anything in that they want (and we're trying to override the props of something somewhere down in the tree). We sort of want to limit them either to a `<Drawer/>` or another component whose first element in the return statement is a `<Drawer/>`. Maybe we can see if it's a drawer, and if not, see if the first `child` is a drawer and clone that like evan suggests?
> 
> @joebochill Do you know if there's a way that we can update the `drawer` prop type to specify that it's either a Drawer or FC that returns a Drawer? I haven't seen anything like this so far.

I know you can type out a function signature in type definitions, but I don't know if it's actually a function that is being passed in. If you do:
```
const MyDrawer = () => <Drawer/>
```
then what is the type of `<MyDrawer/>` in the context of:
```
<DrawerLayout drawer={<MyDrawer/>} />
```
I think it's a JSX.Element, not a function.
```
<DrawerLayout drawer={MyDrawer} />
```
would be a function. Maybe this is the React.Element type?

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Change Log
 
 ## v3.0.0
-- Adds support for nested items in the Drawer component; adjusted Drawer API.
+- Adds support for nested items in the Drawer component; adjusted Drawer and DrawerLayout API.
```
joebochill:
```suggestion
- Adds support for nested items in the Drawer component; 
- Additional styling props added to Drawer
- DrawerLayout API updated for more control over drawer.
```

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@
         ]}
     />
     ```
+- DrawerLayout's `drawer` prop now takes `JSX.Element` instead of `ReactNode`.
```
joebochill: Is this still true

components/src/core/DrawerLayout/DrawerLayout.tsx
```diff
@@ -25,11 +35,41 @@ const useStyles = makeStyles({
...
+        const drawerFC = drawer as React.FC<DrawerProps>;
+        return Object.keys(variantBreakpoints).map(
+            (bp): JSX.Element =>
+                useMediaQuery(bp) &&
+                drawerFC({
+                    key: bp,
+                    variant: variantBreakpoints,
+                    ...drawerElement.props,
+                })
+        );
```
joebochill: I think you need some more control logic here...if the user passes in a JSX.Element, but the displayName is not PXBlueDrawer, it falls back to this FC logic

docs/Drawer.md
```diff
@@ -35,11 +35,17 @@ import { Drawer, DrawerHeader, DrawerSubheader, DrawerBody, DrawerNavGroup, Draw
...
 | width                           | Sets the width of the drawer (in px) when open             | `number ` | no       |          |
 | [...sharedProps](#shared-props) | Props that can be set at any level in the drawer hierarchy | -         | no       |          |
 
 </div>
 
+The `Drawer` has three `variant`s: 
+- **Permanent**: Always open, even when `open` is set to false.
+- **Persistent**: When `open` is set to false, the Drawer collapses itself as a navigation rail, and hover will make it expand temporarily; when `open` is set to true, it behaves like a permanent Drawer.
+- **Temporary**: When `open` is set to false, the drawer is hidden; when `open` is set to true, it slides in.
+The drawer variant overrides that set in variantBreakpoints.
```
joebochill: What is this line saying? If the variant property is set on the drawer, it will override any variantBreakpoint configuration in the DrawerLayout?

docs/Drawer.md
```diff
@@ -282,9 +282,33 @@ The `DrawerLayout` component is used to provide the appropriate resizing behavio
...
+The `variantBreakpoints` prop takes an object, with a media query string as its key, and a Drawer variant as the corresponding value. By default, the prop is set to 
+
+``` typescript
+{
+    [theme.breakpoints.down('xs')]: 'temporary', 
+    [theme.breakpoints.up('sm')]: 'persistent',
+}
+```
+
+which means the Drawer will have variant `temporary` at `xs` or below, and `persistent` at `sm` or above. Manually set the `variant` prop in `<Drawer />` will override this setting.
```
joebochill:
```suggestion
which means the Drawer will have variant `temporary` at `xs` or below, and `persistent` at `sm` or above. Manually setting the `variant` prop in `<Drawer />` will override this setting. 
```

joebochill:
With all of the hassle around this drawer component, I'm inclined to just remove the control logic from the DrawerLayout entirely...it's trivial for an adopter to do:
image
with a couple media queries (and they have to define the media queries anyway if they wanted to use `variantBreakpoints`). It works exactly as expected. We could add a story for this and a blurb in the docs showing how to do a responsive drawer this way.

demos/storybook/stories/drawer/with-different-variants.tsx
joebochill: I would still like to see this in a DrawerLayout...it's hard to understand what the different variants are with the drawer moving around on the screen

PR #64, opened 2020-03-19
Feature/remove storybook warnings

emclaug2:
- StorySource loader does not like it when we export non-story functions; it throws a warning in the console.
- Remove storybook warnings by removing the Icon knob from basic story, and duplicating select knob in all other places.


huayunh:
Thanks Evan! 

@joebochill looks like we'd have to manually duplicate those functions -- storybook does not like it whenever we export some function that returns a JSX.Element (which get recognized automatically as a storybook function return type). So getIcon is recognized as a story but cannot be loaded. Your thoughts on this? 

I am personally open to this -- it actually gives our adopters a better "copiability" when they are on the story knob.

PR #65, opened 2020-03-19
Add Hover State to InfoListItems with onClick Event

### LINKED ISSUE #36: InfoListItem needs a hover state when an onClick is present
#### Describe the bug
No hover state on InfoListItem 

Should show a darker background when hovered and onClick event is defined
		
#### Anything else to add?
This is related to a recent change in the theme, which removed the default hover state. We want to apply this style back, but only when onClick is defined

emclaug2:
Fixes #36 

- Remove most inline styles so that they don't conflict with potential style overrides.
- Add hover state when a click event is present.

### DIFF
--- a/components/src/core/InfoListItem/InfoListItem.styles.tsx
+++ b/components/src/core/InfoListItem/InfoListItem.styles.tsx
@@ -3,6 +3,7 @@ import { InfoListItemProps } from './InfoListItem';
 import color from 'color';
 import * as Colors from '@pxblue/colors';
 
+const getMinHeight = (props: InfoListItemProps): number => props.dense ? 52 : 72;
 const getIconColor = (props: InfoListItemProps): string => {
     const { avatar, iconColor, statusColor } = props;
     if (iconColor) return iconColor;
@@ -15,9 +16,10 @@ const getIconColor = (props: InfoListItemProps): string => {
 export const useStyles = makeStyles<Theme, InfoListItemProps>((theme: Theme) =>
     createStyles({
         root: {
-            backgroundColor: (props): string => props.backgroundColor || 'transparent',
+            backgroundColor: (props): string => props.backgroundColor || theme.palette.background.default,
             cursor: (props): string => (props.onClick ? 'pointer' : 'default'),
-            height: (props): number => (props.dense ? 52 : 72),
+            height: (props): any => (props.wrapSubtitle || props.wrapTitle) ? 'unset' : getMinHeight(props),
+            minHeight: (props): number => getMinHeight(props),
             '&:hover': {
                 backgroundColor: (props): string => (props.onClick ? 'rgba(0, 0, 0, 0.08)' : ''),
             },
@@ -25,9 +27,6 @@ export const useStyles = makeStyles<Theme, InfoListItemProps>((theme: Theme) =>
         avatar: {
             backgroundColor: (props): string => props.statusColor || Colors.black[500],
             color: (props): string => getIconColor(props),
-        },
-        chevron: {
-            display: 'flex',
         },
         divider: {
             position: 'absolute',
--- a/components/src/core/InfoListItem/InfoListItem.styles.tsx
+++ b/components/src/core/InfoListItem/InfoListItem.styles.tsx
@@ -1,0 +1,78 @@
+import { createStyles, makeStyles, Theme } from '@material-ui/core/styles';
+import { InfoListItemProps } from './InfoListItem';
+import color from 'color';
+import * as Colors from '@pxblue/colors';
+
+const getIconColor = (props: InfoListItemProps): string => {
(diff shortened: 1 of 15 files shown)


components/src/core/InfoListItem/InfoListItem.styles.tsx
```diff
@@ -0,0 +1,78 @@
...
+};
+
+export const useStyles = makeStyles<Theme, InfoListItemProps>((theme: Theme) =>
+    createStyles({
+        root: {
+            backgroundColor: (props): string => props.backgroundColor || 'transparent',
+            cursor: (props): string => (props.onClick ? 'pointer' : 'default'),
+            height: (props): number => (props.dense ? 52 : 72),
+            '&:hover': {
+                backgroundColor: (props): string => (props.onClick ? 'rgba(0, 0, 0, 0.08)' : ''),
```
joebochill: how's this work for different themes?

components/src/core/InfoListItem/InfoListItem.test.tsx
```diff
@@ -41,10 +39,11 @@ describe('InfoListItem', () => {
         wrapper = shallow(<InfoListItem hidePadding title="Test" />);
         expect(wrapper.find(PersonIcon).length).toEqual(0);
     });
-
+/* No longer using in-line props to modify styling.
```
joebochill: Can we find another way to test this, or add another test of some sort?

components/src/core/InfoListItem/InfoListItem.tsx
```diff
@@ -98,58 +52,34 @@ export type InfoListItemProps = {
...
                 <ListItemAvatar>
-                    <Avatar
-                        style={{
-                            backgroundColor: statusColor || Colors.black[500],
-                            color: getIconColor(),
-                        }}
-                    >
-                        {icon}
-                    </Avatar>
+                    <Avatar className={defaultClasses.avatar}>{icon}</Avatar>
```
joebochill: if we're moving these into classes, can we exposes entries in the classes prop for users?

components/src/core/InfoListItem/InfoListItem.tsx
```diff
@@ -98,58 +52,34 @@ export type InfoListItemProps = {
...
                 </ListItemAvatar>
             );
         } else if (icon) {
-            return <ListItemIcon style={{ color: getIconColor() }}>{icon}</ListItemIcon>;
+            return <ListItemIcon className={defaultClasses.icon}>{icon}</ListItemIcon>;
```
joebochill: can/should we expose this in classes prop for customization?

components/src/core/InfoListItem/InfoListItem.styles.tsx
emclaug2: This uses the theme that is passed into the MUIThemeProvider.  I tested this out in storybook by setting the InfoListItem's root bg color to use the theme primary, adjusting the light and dark themes to have a difference (light = red, dark = green), then toggling between the themes in the config.js.

components/src/core/InfoListItem/InfoListItem.tsx
emclaug2: I think so, I added it in.

components/src/core/InfoListItem/InfoListItem.tsx
emclaug2: Yep!

components/src/core/InfoListItem/InfoListItem.test.tsx
emclaug2: Absolutely, I'm going to fix this test.  I'll re-request review when the tests are passing.

emclaug2:
> I don't hate breaking out the styles into a separate file to improve readability when there's a lot of them. Not a huge fan of passing around the whole props array to different functions, but it works. You'll have to merge in your changes for the wrapping behavior since that PR went in first.

InfoListItem had about ~80 lines of logic just related to styles and more classes than most other components so I figured it would be a good candidate for a POC.  I like the idea of removing as many of the inline-styles as possible so that they don't conflict with potential style overrides that a user may be passing in.  Now a user's style overrides will always take precedence over our styles or prop inputs.

components/src/core/InfoListItem/InfoListItem.styles.tsx
joebochill: Would it make more sense to have a separate element that goes from transparent to this semi-transparent black color over the existing background?

E.g., assume the user has the background color set to yellow (on a dark gray page). Right now the background would change from `yellow -> almost black` versus `yellow -> dark yellow`.

emclaug2: This was a good catch, I didn't test the hover state on a deep color.  I added a separate element that handles the hover event.

PR #66, opened 2020-03-19
Don't cap max icon size

### LINKED ISSUE #50: Hero height/width doesn't normalize the icon size in the allowable range
#### Describe the bug
The Hero component normalizes the size of the icon itself, but not the height/width of the container, which can lead to odd sizing behavior (see below).

height and width should also be normalized in the allowable range

image

emclaug2:
Fixes #50 

-  Don't cap max icon size at 72px

### DIFF
--- a/components/src/core/Hero/Hero.tsx
+++ b/components/src/core/Hero/Hero.tsx
@@ -53,7 +53,7 @@ const useStyles = makeStyles((theme: Theme) =>
     })
 );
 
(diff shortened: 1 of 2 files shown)


emclaug2:
Capture

emclaug2:
> Need to update the docs surrounding the acceptable range

Good catch, I've updated the docs to specify the min icon size is 10px.

PR #68, opened 2020-03-19
Feature/hero tooltip

### LINKED ISSUE #29: Configurable tooltips on the heros
When the hero text (label, etc.) is too long and ellipses, I want to be able to see the full value in a tooltip.

Text ellipses and there's no way to read it.

This is related to WAS, where many of the channel names are ~40 characters long

#### Additional Context
image

emclaug2:
Fixes #29 

- Add a `enableToolTip` prop on Hero component
- Add a `ToolTipProps` prop on Hero component for any MUI ToolTip overrides
- Add knobs in 'Full Config' story, and 'Within a Banner' Story

### DIFF
--- a/components/src/core/Hero/Hero.tsx
+++ b/components/src/core/Hero/Hero.tsx
@@ -5,7 +5,7 @@ import * as Colors from '@pxblue/colors';
 import { ChannelValue } from '../ChannelValue';
 import PropTypes from 'prop-types';
 import clsx from 'clsx';
-import {Tooltip, TooltipProps as BaseTooltipProps} from '@material-ui/core';
+import { Tooltip, TooltipProps as BaseTooltipProps } from '@material-ui/core';
 
 const useStyles = makeStyles((theme: Theme) =>
     createStyles({
@@ -76,7 +76,7 @@ export type HeroProps = {
     iconSize?: number;
     label: string;
     onClick?: Function;
-    ToolTipProps?: Omit<BaseTooltipProps, 'children'|'title'>;
+    ToolTipProps?: Omit<BaseTooltipProps, 'children' | 'title'>;
     value?: string | number;
     valueIcon?: JSX.Element;
     units?: string;
--- a/components/src/core/Hero/Hero.tsx
+++ b/components/src/core/Hero/Hero.tsx
@@ -1,10 +1,11 @@
-import React from 'react';
+import React, { ReactNode, useCallback } from 'react';
 import { createStyles, makeStyles, Theme } from '@material-ui/core/styles';
 import Typography from '@material-ui/core/Typography';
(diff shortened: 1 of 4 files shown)


emclaug2:
ToolTip Example

emclaug2:
MUI ToolTip override

Here's a picture of what it looks like when using the `ToolTipProps` to override the defaults.

joebochill:
Can you talk to Kyle to see if we have any standard designs for Tooltips (default)?

emclaug2:
> Can you talk to Kyle to see if we have any standard designs for Tooltips (default)?

Kyle said we're using the default MUI tooltip for now, but we might establish some custom defaults when we look further into the ToolTip design pattern.  The default for mobile needs some special attention.

huayunh:
Are we going to provide any special stylings for dark themes?

joebochill:
> > Can you talk to Kyle to see if we have any standard designs for Tooltips (default)?
> 
> Kyle said we're using the default MUI ToolTip for now, but we might establish some custom defaults when we look further into the ToolTip design pattern. The default for mobile needs some special attention.

If we don't have a defined design style around this does it make sense to do it now, or wait until then? Should we be making a separate component (like MUI Tooltip) that we update in one place. Assuming we have tooltip options on all of our components, I don't think we would want to have duplicated code everywhere. 

How do our components work if you just wrap them with a MUI Tooltip component now?

emclaug2:
> > > Can you talk to Kyle to see if we have any standard designs for Tooltips (default)?
> > 
> > 
> > Kyle said we're using the default MUI ToolTip for now, but we might establish some custom defaults when we look further into the ToolTip design pattern. The default for mobile needs some special attention.
> 
> If we don't have a defined design style around this does it make sense to do it now, or wait until then? Should we be making a separate component (like MUI Tooltip) that we update in one place. Assuming we have tooltip options on all of our components, I don't think we would want to have duplicated code everywhere.
> 
> How do our components work if you just wrap them with a MUI Tooltip component now?

If we wrap the entire Hero will the tooltip, the only difference is that the entire Hero would show the tooltip, not just the label.  Right now I only wrapped the label part of the Hero because that's the element that commonly overflows.  If you want to wrap the entire Hero with a tooltip, we shouldn't be adding this functionality to the Hero component; just wrap the Hero with a ToolTip.

We should only make a separate PX Blue ToolTip if there are enough customizations to justify one.  Regardless of whether we use the MUI or a wrapper for the tooltip, the Hero API should remain unchanged since we accept ToolTipProps at the Hero level.

Also, I don't think our PX Blue themes have any ToolTip styles.

emclaug2:
Closing this because this might be an unnecessary feature.   There is a branch 'experiment/hero-tooltip' that can referenced to.

PR #73, opened 2020-03-23
Feature/usermenu support nonclickables

### LINKED ISSUE #38: UserMenu to support non-clickable items
I want UserMenu to support non-clickable items.
If no onClick is passed to a UserMenuItem, cursor should be text and onClick should not dismiss the Menu.

This was requested by Anjali who had a use-case of wanting a read-only row at the bottom of the menu.  
All UserMenu items have a cursor pointer and dismiss the Menu when clicked, even if there is no click event passed onto the item. 

No

### LINKED ISSUE #71: InfoListItemProps on DrawerNavGroup, DrawerBody, etc
Add the ability for a user to pass in props directly to the underlying InfoListItems inside of the Drawer.  This would allow a user ability to override styles inside of the InfoListItem component.

No such spreading.  Can't override any props. 

No

emclaug2:
Fixes #71 
Fixes #38 

- Add InfoListItemProps to Drawer Inheritable Props
- Separate DrawerNavItem from DrawerNavGroup file.
- Add ability to make UserMenu contain items non-clickable items that don't close menu onClick.
- Add example to Storybook.

### DIFF
--- a/components/src/core/Drawer/DrawerNavItem.tsx
+++ b/components/src/core/Drawer/DrawerNavItem.tsx
@@ -1,6 +1,6 @@
-import React  from 'react';
+import React from 'react';
 import { createStyles, makeStyles, Theme, useTheme } from '@material-ui/core/styles';
-import { ArrowDropUp, ChevronRight, ExpandLess } from '@material-ui/icons';
+import { ArrowDropUp, ChevronRight, ExpandLess, Info } from '@material-ui/icons';
 import { gray } from '@pxblue/colors';
 import clsx from 'clsx';
 import { PXBlueDrawerInheritableProperties } from './Drawer';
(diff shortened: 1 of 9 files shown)


components/src/core/UserMenu/UserMenu.tsx
```diff
@@ -62,6 +62,29 @@ export const UserMenu: React.FC<UserMenuProps> = (props) => {
...
+        for (const group of menuGroups) {
+            for (const item of group.items) {
+                const onClick = item.onClick;
+                if (onClick) {
+                    item.onClick = (): void => {
+                        onClick();
+                        closeMenu();
+                    }
+                } else {
+                    item.InfoListItemProps = { ripple: false, ...item.InfoListItemProps };
```
joebochill: should we instead update the InfoListItem to not show the ripple unless there is a onClick event? Like we do with hover?

components/src/core/Drawer/DrawerNavItem.tsx
```diff
@@ -0,0 +1,232 @@
...
+                        </div>
+                    )
+                }
+                backgroundColor={'transparent'}
+                onClick={hasAction ? onClickAction : undefined}
+                style={{ paddingLeft }}
+                hidePadding={hidePadding}
+                ripple={ripple}
+                classes={depth > 0 ? { title: clsx(defaultClasses.nestedTitle, classes.nestedTitle) } : {}}
+                {...InfoListItemProps}
```
joebochill: we should merge our classes with their classes...otherwise if they want to put a class on something other than the nested items, it will blow away our nested class/styles.

components/src/core/UserMenu/UserMenu.tsx
emclaug2: I agree, I made the MUI ListItem `button` prop dependent on the the InfoListItem's `ripple` and `onClick` being set.   It doesn't make sense for non-clickable items to have a ripple effect for now reason, right?

components/src/core/Drawer/DrawerNavItem.tsx
emclaug2: I added a fix to combine the classes for nested Items.  If they pass in a class to the InfoListItem title, it wont be applied to the nested items.

PR #84, opened 2020-03-26
Remove the div wrapper around ILI

joebochill:
- Remove the extra div wrapper around the InfoListItem
- Use colors package to darken background on hover instead of extra dom element


emclaug2:
This is the current Landing Page with the Open Sans applied.   
Open Sans

This is what currently is in the Dev Branch.
React-Dev

emclaug2:
This is what the READMEs look like with the Open Sans applied.  I think adjusting the @pxblue/storybook-themes to have a lesser H1 font weight would be nice here.
Open Sans README

This is what the current READMEs look like in dev.
React-Dev README

PR #85, opened 2020-03-26
Apply open sans to storybook app, change bg image size

emclaug2:
- Apply Open Sans font to Storybook
- Adjust welcome page margin so content is centered.
- Remove bg size of 200% (?)

### DIFF
--- a/demos/storybook/.storybook/manager-head.html
+++ b/demos/storybook/.storybook/manager-head.html
@@ -17,4 +17,134 @@
     .css-cz6gk6 {
         color: #007bc1 !important; /* Blue[500] */
     }
+
+    /* open-sans-300normal - latin */
+    @font-face {
+        font-family: 'Open Sans';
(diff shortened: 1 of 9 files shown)


emclaug2:
This is the current Landing Page with the Open Sans applied.   When I look at the background color from up close, I  like the smaller circles.  Whenever I see it from afar, I'm a fan of the larger background.   Curious what other people think.
Open Sans

This is what currently is in the Dev Branch.
React-Dev

emclaug2:
This is what the READMEs look like with the Open Sans applied.  I think adjusting the @pxblue/storybook-themes to have a lesser H1 font weight would be nice here.
Open Sans README

This is what the current READMEs look like in dev.
React-Dev README

emclaug2:
This is what I want the READMEs to look like with the heavy H1 font weight removed: 
image

PR #89, opened 2020-03-28
Update component docs

huayunh:
- Add infographics for the UserMenu and Hero; updated the screenshot of the InfoListItem
- Correct two wrong documentations: 
  - `UserMenu.menuGroups` accepts `UserMenuGroups[]` instead of just `UserMenuGroups`
  - `InfoListItem.wrapTitle` and `InfoListItem.wrapSubtitle` should be false by default (marked as true)


huayunh:
Interesting... CircleCI just replenished our org the new quota?

docs/UserMenu.md
```diff
@@ -54,14 +57,14 @@ const avatar = <Avatar><SendIcon/></Avatar>;
 |-------------------|--------------------------------------------------|--------------------------|----------|-------------|
 | avatar            | MUI Avatar that displays                         | `Avatar`                 | yes      |             |  
 | classes           | Style overrides                                  | `StyleRules`             | no       |             |           
-| onClose           | Function called when the menu is closed          | `Function`               | no       |             |                                         
-| onOpen            | Function called when the menu is opened          | `Function`               | no       |             |                                         
 | menu              | Custom MUI Menu displayed when Avatar is clicked | Material-UI `Menu`       | no       |             |                                         
-| menuTitle         | Title shown when menu is open                    | `string`                 | no       |             |                                         
+| menuGroups        | Groups of menu items that display                | `UserMenuGroups[]`       | yes      |             |
```
joebochill: menuGroups is not required

huayunh: `menuGroups` is required in PropType but not required in `UserMenuProps` type. A user menu with no menuGroups will yield only a single avatar that does no harm to the rest of the world.

I personally prefer it to be required to make this component meaningful.

huayunh: Although this will theoretically break some code and make us publish react-component-library4.0.0 after only one weekend?

joebochill: menuGroups is not required because you can pass in your own custom `Menu` component. Fixing the PropTypes won't necessitate a 4.0.0

huayunh: I see, that makes sense.

PR #96, opened 2020-04-01
Remove references to palette weights

joebochill:
- Removes all references the `theme.palette.<color>[weight]`
- Uses lighten function from MUI to get the [50] shade of the primary palette main


huayunh:
What was the reason to use `lighten` instead of setting `theme.palette.primary.light` to blue50? I am checking DSM and see that blue 50 is used for all of these, but didn't see any place where blue300 is used in DSM. Blue50 seems to be the most common light blue background: 

## disabled button
image

## menu items
image

## date picker
image

joebochill:
@huayunh let me check the theme demo and see if 50 will break anything

joebochill:
> @huayunh let me check the theme demo and see if 50 will break anything

My only concern with that would be if Material UI starts using `theme.palette.light` to style their components (they're using 300/700 for light/dark in their palettes). They currently use `theme.palette.dark` in a couple of their components, but so far they don't use light.

PR #118, opened 2020-04-28
Update inline styles to classes & extend top-level component props

huayunh:
Fixes **PXBLUE-1212** **PXBLUE-1215**

- Removed "inline styles (and add classes), similar to InfoListItem so that class overrides always take precedence. "
- Made "sure that every component has a 'classes' prop for further customization"
- Extended the props to introduce the top-level component props for each component
  - I did not change DrawerNavItem due to the unique way it is implemented (it has an `InfoListItemProps` prop already)
  - Altered the `HeroBanner` FC to take `props: HeroBannerProps` rather than `props: HeroBannerProps & any`
  - For things wrapped in a `React.Fragment`(such as hero banner), I didn't interpret the fragment as the top-level component. Only the ones marked as "root"'s are considered the top-level. 
- Changelog updated. Version updated to 3.1.0


huayunh:
I am too afraid of the merge conflicts, so marking as a draft to do two stories at once

PR #136, opened 2020-06-08
Update to consistent usage of vertical bar in readme tables

joebochill:
For PXBLUE-1355 .

- Update to consistent usage of vertical bar in readme tables


joebochill:
> So essentially just adding extra spaces?

In this case, yes :)

PR #147, opened 2020-06-24
Add support for a third line of text on the InfoListItem

### LINKED ISSUE #3: InfoListItem should support third line of text
Material Design Lists support up to three lines of text in list items

InfoListItem only allows configuration for title and subtitle (2 lines)

No

#### Additional Context
This will need to be updated for the React and React Native libraries. Ask Eric for style input.

daileytj:
Fixes #3.

- Add 'info' prop to InfoListItem to add support for a third line of text.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## V4.0.2
+
+- Add `info` prop to InfoListItem to support a third line of text.
+
 ## v4.0.1
 
 -   Themes updated to use [@pxblue/colors](https://www.npmjs.com/package/@pxblue/colors) version 3.0.0.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## V4.0.2
 
-- Add `info` prop to InfoListItem to support a third line of text.
+- Add `info` prop to `<InfoListItem>` to support a third line of text.
 
 ## v4.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## V4.0.2
 
 - Adds `info` prop to `<InfoListItem>` to support a third line of text.
-- Change title and description types from `string` to `ReactNode`.
+- Change title and description types from `string` to `ReactNode` for `<EmptyState>`.
 
 ## v4.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## V4.0.2
 
-- Add `info` prop to `<InfoListItem>` to support a third line of text.
+- Adds `info` prop to `<InfoListItem>` to support a third line of text.
 
 ## v4.0.1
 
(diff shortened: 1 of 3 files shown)


components/src/core/InfoListItem/InfoListItem.tsx
```diff
@@ -36,6 +36,7 @@ export type InfoListItemProps = Omit<Omit<ListItemProps, 'title'>, 'divider'> &
     statusColor?: string;
     subtitle?: string | Array<string | JSX.Element>;
     subtitleSeparator?: string;
+    info?: string | Array<string | JSX.Element>;
```
huayunh, line 39: Should these `<string | JSX.Element>` be just ReactNodes? @joebochill

joebochill: Tom and I discussed this a little earlier. I think there's a bigger refactor coming for the subtitle where we will just accept a `ReactNode` and get rid of all of the internal logic for adding in the separator. Since that would be a breaking change, we decided to leave it as is for the moment and let the `info` prop match the implementation/usage of subtitle.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## V4.0.2
+
+- Add `info` prop to InfoListItem to support a third line of text.
```
joebochill, line 5: You'll need to merge this into the other changes for the changelog...just merge with dev branch:
```suggestion
- Adds `info` prop to `<InfoListItem>` to support a third line of text.
```

components/src/core/InfoListItem/InfoListItem.tsx
```diff
@@ -129,7 +145,12 @@ export const InfoListItem: React.FC<InfoListItemProps> = (props) => {
             <ListItemText
                 primary={title}
                 className={combine('listItemText')}
-                secondary={getSubtitle()}
+                secondary={
+                    <>
+                        <Typography variant={'body2'}>{getSubtitle()}</Typography>
+                        <Typography variant={'body2'}>{getInfo()}</Typography>
+                    </>
```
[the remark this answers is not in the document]
daileytj: Looks fine. If you don't pass in either a subtitle or info then they don't render, which does mean, however, that if you use the info prop without a subtitle then it will look just like the subtitle

huayunh:
Tom you might want to do as many commits as you can locally before git push, just to save up some ci quota (and trigger less GitHub notifications)

daileytj:
> Tom you might want to do as many commits as you can locally before git push, just to save up some ci quota (and trigger less GitHub notifications)

I'll be sure to keep an eye on that. Thanks

joebochill:
Just a thought, but do we want to hide the `info` line if the `dense={true}`. I'm thinking of how the Header works for React Native...when it's collapsed the info line fades out. But maybe it's better not to have that logic in here for now and just leave it up to the user.

daileytj:
> Just a thought, but do we want to hide the `info` line if the `dense={true}`. I'm thinking of how the Header works for React Native...when it's collapsed the info line fades out. But maybe it's better not to have that logic in here for now and just leave it up to the user.

Up to you, just let me know what you want to do.

joebochill:
> > Just a thought, but do we want to hide the `info` line if the `dense={true}`. I'm thinking of how the Header works for React Native...when it's collapsed the info line fades out. But maybe it's better not to have that logic in here for now and just leave it up to the user.
> 
> Up to you, just let me know what you want to do.

I think we can leave it out for now, just wanted to capture the thought.

PR #152, opened 2020-06-26
Use prettier check

huayunh:
- Add prettier:check
- 
-


huayunh:
I see what's going on here

joebochill:
> I see what's going on here

I've run into this locally a bunch of times...very annoying

PR #158, opened 2020-07-08
Add RTL wrapper for showcase

joebochill:
- Adds a RTLProvider component to easily enable/disable RTL in the showcase (default off)
- Fix style prop spreading in the Drawer component
- Switch to class overrides in the UserMenu component (outstanding TODO)

The only component that seems to have any significant RTL issues is the Drawer/DrawerLayout since it is currently hardcoding a padddingLeft value in the style (via JS), which is due for a little refactoring anyway.

#### Screenshots:
image


emclaug2:
The wrapper works, our DrawerLayout def needs some work.

This is what I see: 
image

huayunh:
You said the default is off for RTL? The showcase is giving me RTL on yarn start...

huayunh:
Theme toggle is not working

huayunh:
Is user menu supposed to be the way they are right now? I see that the list items & header have their icons / avatars on the left under RTL. Or will this be covered in the i18n component refactor?

huayunh:
image

use `startIcon` for `<Add>` instead. Or `endIcon` if still having issues with the icon padding

joebochill:
@huayunh good catches. I changed the default and did a little refactoring to add a toggle switch in the toolbar for RTL/LTR. I also fixed the UserMenu (the RTL provider was nested further down in the DOM tree, so items higher in the tree, such as overlays, weren't getting the new direction - I reconfigured the RTL provider to swap the direction on the body tag instead). Other application-side changes should be handled in the i18n update story...this was mostly just a convenience to be able to quickly see what RTL looks like in its current state.

PR #163, opened 2020-07-14
Fix issue with ILI subtitle wrapping

### LINKED ISSUE #160: InfoListItem not wrapped or clipped properly
#### Describe the bug
InfoListItem used to clip an overflowing text or wrap it, now the component doesn't do that. (default behavior is an overflow:hidden)

When there is a long text, either wrap the title or clip it with an ellipsis 

#### What are the steps to reproduce?
1. Go to storybook
2. shrink the canvas so that the text overflow
3. observe that the text just gets clipped off

image
from https://pxblue-components.github.io/react/?path=/story/components-info-list-item--with-right-component

#### Environment
Chrome on MacOS
		
#### Anything else to add?
wrapTitle & wrapSubtitle is still working just fine.

joebochill:
Fixes #160 

- Apply the noWrap property to each of the subtitle and info typography elements
- Update the readme description of fontColor prop

#### Screenshots:
image

### DIFF
--- a/components/src/core/InfoListItem/InfoListItem.styles.tsx
+++ b/components/src/core/InfoListItem/InfoListItem.styles.tsx
@@ -8,6 +8,7 @@ export type InfoListItemClasses = {
     avatar?: string;
     divider?: string;
     icon?: string;
+    info?: string;
     listItemText?: string;
     rightComponent?: string;
     separator?: string;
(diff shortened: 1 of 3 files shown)


emclaug2:
Is `noWrap` still needed on the primary and secondary typography props?

joebochill:
> Is `noWrap` still needed on the primary and secondary typography props?

It was no longer required on the SecondaryTypographyProps, so I removed it. I also added a style class for the info line and moved the subtitle one where it needed to be to work correctly.

PR #165, opened 2020-07-21
Update drawer padding for RTL, fix showcase styles

joebochill:
For PXBLUE-1456.

- Update the drawer component to conditionally do left/right padding on the DrawerLayout based on theme direction
- Update the showcase demo to move styles into JSS classes
    - This lets the JSS plugin automatically flip the styles when we go to RTL
- Add logic to flip all of the icons (Except for GradeA since it has text in it, not that it really makes a difference)
    - The Pie progress icon does't flip because of a bug (https://github.com/pxblue/icons/issues/65)
- Fixes a bug in the ILI hover color (doesn't work when using rgba color for background)
    - Reported by Anjali today

#### Screenshots:
image

> I'm still looking into an alternative for linking the DrawerLayout and the Drawer other than doing the lookup by ID, but this works for now keeping the existing approach.

### DIFF
--- a/demos/showcase/src/App.js
+++ b/demos/showcase/src/App.js
@@ -23,6 +23,7 @@ import {
 } from '@pxblue/react-components';
 
 import top from './topology_40.png';
+import clsx from 'clsx';
 
 const useStyles = makeStyles((theme) =>
     createStyles({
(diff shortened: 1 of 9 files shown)


demos/showcase/src/App.js
```diff
@@ -262,6 +278,11 @@ export const App = () => {
                         />
                     </List>
                 </Card>
+                {/* <DrawerLayout
```
emclaug2, line 281: Should this be removed?

emclaug2:
This looks fantastic!

jeffvg:
this looks good. Did we talk about this?

image

joebochill:
> this looks good. Did we talk about this?
> 
> image

This is interesting. It seems there is some browser smartness around unicode characters like the degree symbol to move them to the other side when doing RTL (cool!). In practice, users would actually be using a RTL language when the flip to the RTL layout and the text would flow more naturally than it looks here in English.

PR #166, opened 2020-07-22
Refactor drawerLayout to use ID and React context

joebochill:
For PXBLUE-1456.
Fixes #41 

RefactorRefactor the Drawer/DrawerLayout to link them in the case when multiple drawers are used
- By default, a Drawer will be linked to it's closest DrawerLayout ancestor (without any changes in usage)
    - This also means that if you are using multiple DrawerLayouts with Drawers, they will work fine without any changes in usage (each Drawer will simply operate the DrawerLayout closest to it in the DOM tree)
- ~~The only case where you need to use the `layoutID` prop on Drawer and DrawerLayout is if you have some Drawers inside a DrawerLayout and some without. In order to prevent the rogue drawer from affecting the nearest DrawerLayout, you'll need to give the other ones the ID.~~
- The only case where a use needs to change anything is if they are using multiple Drawers and some of them are inside of a DrawerLayout and some are not. In this case, they will need to identify the Drawers that are not part of a DrawerLayout by adding the `noLayout` prop.
    - This is a very very small use-case, 99% of the time nobody is going to run into this, so they'll be able to continue using the Drawer/DrawerLayout as before without needing to make any changes.
- Updated documentation

The following should all work as expected:
```
<DrawerLayout drawer={<Drawer />} />
```
```
<DrawerLayout drawer={<Drawer />} />
...
<DrawerLayout drawer={<Drawer />} />
```
```
<Drawer />
```
```
<DrawerLayout drawer={<Drawer />}>
    <Drawer noLayout/>
</DrawerLayout>
```

### DIFF
--- a/components/src/core/Drawer/Drawer.tsx
+++ b/components/src/core/Drawer/Drawer.tsx
@@ -118,7 +118,7 @@ export const DrawerComponent: React.FC<DrawerComponentProps> = (props) => {
     let hoverDelay: NodeJS.Timeout;
     const defaultClasses = useStyles(props);
     const theme = useTheme();
-    const {onPaddingChange} = useDrawerLayout();
+    const { onPaddingChange } = useDrawerLayout();
     const [hover, setHover] = useState(false);
     const {
         activeItem,
@@ -264,7 +264,7 @@ export const DrawerComponent: React.FC<DrawerComponentProps> = (props) => {
     const contentWidth = width || defaultContentWidth;
 
     useEffect(() => {
-         onPaddingChange(layoutID, variant === 'temporary' ? 0 : containerWidth);
+        onPaddingChange(layoutID, variant === 'temporary' ? 0 : containerWidth);
     }, [containerWidth, variant, theme]);
 
     return (
--- a/components/src/core/Drawer/Drawer.tsx
+++ b/components/src/core/Drawer/Drawer.tsx
@@ -103,11 +103,12 @@ export type PXBlueDrawerNavGroupInheritableProperties = {
 
 export type DrawerComponentProps = {
     classes?: DrawerClasses;
+
+    // Describes if this Drawer is used outside of a DrawerLayout
+    noLayout?: boolean;
+
     // Controls the open/closed state of the drawer
     open: boolean;
-
-    // Refers to the id of the parent DrawerLayout
-    layoutID?: number | string;
 
     // Sets the width of the drawer (in px) when open
     width?: number;
@@ -135,9 +136,9 @@ export const DrawerComponent: React.FC<DrawerComponentProps> = (props) => {
         InfoListItemProps,
         itemFontColor,
         itemIconColor,
-        layoutID,
         nestedBackgroundColor,
         nestedDivider,
+        noLayout = false,
         open,
         onItemSelect,
         ripple,
@@ -264,8 +265,8 @@ export const DrawerComponent: React.FC<DrawerComponentProps> = (props) => {
     const contentWidth = width || defaultContentWidth;
 
     useEffect(() => {
-        onPaddingChange(layoutID, variant === 'temporary' ? 0 : containerWidth);
-    }, [containerWidth, variant, theme]);
+        if (!noLayout) onPaddingChange(variant === 'temporary' ? 0 : containerWidth);
+    }, [containerWidth, variant, noLayout]);
 
     return (
         <Drawer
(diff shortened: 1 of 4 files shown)


joebochill:
#### Question
Do we want to go this approach where we provide an ID on the drawer and the drawer layout that need to match, or would we prefer to have a boolean prop on the drawer that identifies it as a rogue drawer:
```
<Drawer noLayout />
```

If we were to take this approach, the end result would be the same. One benefit of the other approach is that you can arbitrarily link a drawer to any DrawerLayout in the application (not sure if this is something we want to encourage or even support). Thoughts?

joebochill:
I'm kind of leaning towards switching to the boolean approach since it would only require users to make a change in one place. It also would ensure that drawers are only affecting their parent layout and no others.

emclaug2:
> ## Question
> If we were to take this approach, the end result would be the same. One benefit of the other approach is that you can arbitrarily link a drawer to any DrawerLayout in the application (not sure if this is something we want to encourage or even allow). Thoughts?

This seems unnecessary; I'm having a hard time imaging a situation where DrawerA would want to affect DrawerLayoutB.   I think it's good to just scrap that functionality all together.

I really like how flexible this patch is in terms of compatibility; and the boolean approach seems fine to me.  Most users will never have to be concerned with these changes and this is a good thing.

joebochill:
I went ahead and switched the implementation over to the boolean prop on the drawer approach...I think it feels cleaner. 🧻

huayunh:
When the drawer variant is `temporary`, it looks good. 

However, if I try doing `permanent` or `persistent`, here's what's happening:

If I change `<App />` to:
```jsx
<TheOuterBigDiv>
    <SecondaryDrawer />
    ...other stuff
</TheOuterBigDiv>
```

image

or if I change `<App />` to:

```html
<>
    <SecondaryDrawer />
    <TheOuterBigDiv />
</>
```
then it looks like
image
(has to scroll down 100vh to see the contents from `<TheOuterBigDiv />`)

The drawer I am using looks like this:
```jsx
export const SecondaryDrawer = (props) => {
    return (
        <PXBDrawer open={true} variant={'permanent'} noLayout {...props}>
            <DrawerHeader title={'secondary drawer'} />
            <DrawerBody>
                <DrawerNavGroup
                    items={[
                        { title: 'test', itemID: 'test', icon: <SendIcon /> },
                        { title: 'test2', itemID: 'test2', icon: <SendIcon /> },
                    ]}
                />
            </DrawerBody>
        </PXBDrawer>
    );
};
```

joebochill:
@huayunh what you're describing is part of the current nature of the `<Drawer>` when it is used outside of the DrawerLayout (it's actually one of the main reasons why we recommend using the DrawerLayout). It uses vertical flexBox to lay out its content and so if it doesn't have an explicit size specified, it will get condensed like you're seeing. If you do a little flex-boxing you can get it to look as you'd expect:
image

You'll see the same behavior if you just try to use a Drawer on its own (without the rest of the application). I actually built a quick demo of this for somebody who sent us an email with the same issue: https://codesandbox.io/s/nifty-swirles-rnjkb?file=/src/WithoutDrawerLayout.js

joebochill:
I implemented the secondary drawer in the router like this:
```tsx
export const MainRouter = () => {
    const [open, setOpen] = useState(true);
    const [open2, setOpen2] = useState(true);
    return (
        <Router>
            <ScrollToTop />
            <DrawerLayout drawer={<NavigationDrawer open={open} setOpen={setOpen} />}>
                <SharedAppBar
                    onClick={() => {
                        setOpen(!open);
                    }}
                />
                <div style={{ display: 'flex' }}>
                    <PXBDrawer open={open2} variant={'temporary'} noLayout onClose={() => setOpen2(false)}>
                        <DrawerHeader title={'secondary drawer'} onClick={() => setOpen2(!open2)} />
                        <DrawerBody>
                            <DrawerNavGroup
                                items={[
                                    { title: 'test', itemID: 'test' },
                                    { title: 'test2', itemID: 'test2' },
                                ]}
                            />
                        </DrawerBody>
                    </PXBDrawer>
                    <div style={{ flex: '1 1 0px' }}>
                        <Switch >
                            <Route path="*">
                                <App />
                            </Route>
                        </Switch>
                    </div>

                </div>
            </DrawerLayout>
        </Router>
    );
};
```

huayunh:
I see. So I guess no 
```jsx
<DrawerLayout drawer={<Drawer />}>
    <DrawerLayout drawer={<Drawer />} />
</DrawerLayout>
```

Sounds good to me

joebochill:
> I see. So I guess no
> 
> ```js-jsx
> <DrawerLayout drawer={<Drawer />}>
>     <DrawerLayout drawer={<Drawer />} />
> </DrawerLayout>
> ```
> 
> Sounds good to me

Logically, I think that should work fine. Making it look right might be another story (if we're doing anything like 100vh). I suppose you could do something like that to make like a mini 'iframe' type thing in your app if you wanted to...maybe.

PR #173, opened 2020-07-31
Feature/rtl addon

### LINKED ISSUE #172: RTL Addon for Storybook
Add a RTL toggle to our storybook app.

No such toggle exists.

No

emclaug2:
Fixes #172 

- Add RTL addon to storybook

### DIFF
--- a/demos/storybook/.storybook/config.js
+++ b/demos/storybook/.storybook/config.js
@@ -88,11 +88,11 @@ addDecorator((storyFn) => {
     const [darkTheme, setDarkTheme] = useState(createMuiTheme(appendDirection(ReactThemeDark)));
 
     useEffect(() => {
-        channel.on(DIR_CHANGE_EVENT, update);
-        return () => channel.off(DIR_CHANGE_EVENT, update);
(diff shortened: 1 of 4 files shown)


emclaug2:
Any RTL defects found in individual components will require separate issues to be opened.

emclaug2:
Firebase deployment for easier viewing:  https://pxblue-react-library.web.app/

jeffvg:
this is to only throw in the add-on piece and not fix anything but to log issues later?

emclaug2:
> this is to only throw in the add-on piece and not fix anything but to log issues later?

That's correct - if there's anything wrong with the RTL version in any of our components, the fix should be logged as a separate issue and not fixed inside this PR.  This is just to prevent scope creeping.

emclaug2:
For example, none of the icons are inverted inside a hero component when RTL is enabled.

joebochill:
> None of the icons are inverted inside a hero component when RTL is enabled. I'll make an issue for this one.

None (90%) of the icons won't be flipped because our react components don't do automatic flipping (except for internally used icons)...Any icons passed into components will need to be flipped on the storybook side.

PR #174, opened 2020-08-04
Feature/showcase submodule

joebochill:
For PXBLUE-1434 .

- Replace showcase with submodule ([react-showcase-demo](https://github.com/pxblue/react-showcase-demo))
- Convert the showcase to typescript and add linting configuration
- Refactor some of the showcase redux integrations to use hooks
- Added new commands to the component library
    - `yarn initialize` initializes the submodule (this will automatically be run as part of the existing start command)
    - `yarn update:submodule` should be run if there are upstream changes to the showcase repo that you need to pull down


emclaug2:
Looks like circleci needs the full git url when attempting to clone the submodule.

emclaug2:
I ran `yarn start:showcase` and ran into an infinite loop.

Is this command supposed to work or should it be removed?

joebochill:
> I ran `yarn start:showcase` and ran into an infinite loop.
> 
> Is this command supposed to work or should it be removed?

The command will still work, but after initial clone, there is an additional step that must be run `yarn initialize` to clone the contents of the submodule (otherwise the start command in that folder will execute the start command of the component repo, which causes a circular loop like you're seeing.

I added a new script to conditionally do the initialization automatically, so now `yarn start:showcase` should work off of a fresh clone the way you're used to.

PR #176, opened 2020-08-06
Add optional divider prop to DrawerHeader, default false

### LINKED ISSUE #164: Turn off the Drawer Header divider by default

Would like to have drawer header divider turned off by default, and turned on only when a new prop `divider` is set to true

image
From docit. This alignment is awkward

Can be part of the drawer refactor

emclaug2:
Fixes #164 

- Divider under DrawerHeader is now optional and by default turned off.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Fixes RTL support issues in the `<DrawerLayout>`/`<Drawer>`
 - Fixes an issue with text wrapping in the `<InfoListItem>` subtitle
 - Fixes an issue with hover color alpha channel in `<InfoListItem>`
+- Adds optional prop to the `<DrawerHeader>` to add a bottom divider
 
 ## V4.0.2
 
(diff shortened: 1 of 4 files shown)


huayunh:
Just a note: the header divider is off by default given our current white on blue header design; when we reach a consensus on the drawer style, it _might_ change to default on.

image
(I know Joe is still unhappy about this design decision)

joebochill:
> Just a note: the header divider is off by default given our current white on blue header design; when we reach a consensus on the drawer style, it _might_ change to default on.
> (I know Joe is still unhappy about this design decision)

I'm fine with showing white headers in storybook, I just wouldn't make them the default.

I'm fine with this change as is though, since with the blue default you don't really see the divider anyway

huayunh:
And update storybook full config story?

emclaug2:
I've updated the changelog and the full config story.  

Also having fun with this gif maker. ++
Drawer no icon

emclaug2:
I've updated the full config story and the changelog.

PR #177, opened 2020-08-06
Icon align prop

huayunh:
Fixes 1472.

- Added `iconAlign` prop (like `textAlign`) that aligns the icon when `avatar` prop is set to `false`. This prop is set to `left` by default
- Added story, doc, changlog... marking as a draft until I figure out how to do the git submodule thing

#### Screenshots:

Under LTR and dark mode.
Note that the drawer menu items did not get any settings on it; the only things changed in this screenshot are those in the list of info list items. Look how nicely aligned they are :) 

image


components/src/core/InfoListItem/InfoListItem.test.tsx
```diff
@@ -61,12 +59,6 @@ describe('InfoListItem', () => {
     });
  */
 
-    it('renders with avatar', () => {
```
joebochill, line 64: No love for the avatar test?

huayunh: Yup, bc now the thing renders a transparent avatar background when avatar = false

huayunh:
> Works great! My only question is do we want these to be 'left' 'center' 'right', or 'start' 'center' 'end'? I only ask because i looked at it in RTL and it was backwards to what the prop says (technically).

Should I do `'left' | 'start' | 'center' | 'right' | 'end'` then? :P 

I am happy with either way -- we are having "left component" "right component" etc, so I think it should be fine to leave it that way

joebochill:
> > Works great! My only question is do we want these to be 'left' 'center' 'right', or 'start' 'center' 'end'? I only ask because i looked at it in RTL and it was backwards to what the prop says (technically).
> 
> Should I do `'left' | 'start' | 'center' | 'right' | 'end'` then? :P
> 
> I am happy with either way -- we are having "left component" "right component" etc, so I think it should be fine to leave it that way

Good point...I'm fine with it as is

PR #188, opened 2020-08-23
Hide title & subtitle on persistent drawer collapse

### LINKED ISSUE #178: Persistent Drawer hide labels on collapse
#### Describe the bug
The Drawer doesn't look good when used with no icon and collapsed in the persistent variant. 

It should hide the text in the nav items when collapsed, similar to the angular component library.

React Drawer Bug

huayunh:
Fixes #178 

- Hide title & subtitle upon the persistent drawer collapse
- Should only work when you have (1) `hidePadding` set and (2) no `icon` passed in
- Easiest hack... without getting the code too messy?

#### Screenshots:
ezgif-4-b8240779222d


CHANGELOG.md
```diff
@@ -1,12 +1,18 @@
...
+-   Adds optional prop to the `<DrawerHeader>` to add a bottom divider
+
+### Fixed
+
 -   Fixes RTL support issues in the `<DrawerLayout>`/`<Drawer>`
 -   Fixes an issue with text wrapping in the `<InfoListItem>` subtitle
 -   Fixes an issue with hover color alpha channel in `<InfoListItem>`
--   Adds `iconAlign` prop to `<InfoListItem>` to align icon left (default), center or right
--   Adds optional prop to the `<DrawerHeader>` to add a bottom divider
+-   Hides Drawer Nav Item's title when the `<Drawer>` collapses in the `persistent` variant
```
joebochill, line 15: I would categorize this one as a 'Changed' instead of a 'Fixed'

components/src/core/Drawer/DrawerNavItem.tsx
```diff
@@ -196,7 +206,12 @@ export const DrawerNavItem: React.FC<DrawerNavItem> = (props) => {
     const active = activeItem === itemID;
     const infoListItemClasses = {
         root: defaultClasses.infoListItem,
-        title: depth > 0 ? clsx(defaultClasses.nestedTitle, classes.nestedTitle) : '',
+        title: clsx(
+            depth > 0 && defaultClasses.nestedTitle,
+            depth > 0 && classes.nestedTitle,
+            hidePadding && !icon && defaultClasses.noIconTitle
+        ),
+        subtitle: clsx(hidePadding && !icon && defaultClasses.noIconTitle),
```
joebochill, line 214: clsx has a nice syntax for conditionally including classes:
```suggestion
        title: clsx({
            [defaultClasses.nestedTitle]: depth >0,
            [classes.nestedTitle]: depth > 0,
            [defaultClasses.noIconTitle]: hidePadding && !icon,
        }),
        subtitle: clsx({
            [defaultClasses.noIconTitle]: hidePadding && !icon,
        }),
```

components/src/core/Drawer/Drawer.tsx
```diff
@@ -283,7 +283,15 @@ export const DrawerComponent: React.FC<DrawerComponentProps> = (props) => {
                 drawerProps.style
             )}
         >
-            <div className={clsx(defaultClasses.content, classes.content)} style={{ width: contentWidth }}>
+            <div
+                className={clsx(
+                    defaultClasses.content,
+                    classes.content,
+                    'pxb-drawer-content',
+                    isDrawerOpen() && 'pxb-drawer-open'
```
joebochill, line 291: Rather than passing around these non-JSS class names, I would recommend adding a React Context around the drawer that makes some drawer context available to its various children (like the DrawerNavItem), mostly just the open state right now, but we could possibly use it for some of the prop inheritance logic too down the road.

If you look at the Header component in RN, we use a couple of contexts to pass information down this way (such as the height of the animated header): https://github.com/pxblue/react-native-component-library/blob/dev/components/src/core/header/header.tsx#L287

The context provider is defined here: https://github.com/pxblue/react-native-component-library/blob/dev/components/src/core/header/contexts/HeaderHeightContextProvider.tsx

You could then just conditionally add your noIconTitle class below based on:
```
[defaultClasses.noIconTitle]: hidePadding && !icon && !drawerOpen
```

When setting up the context, you'll want to make sure that if context === null, you'll just want to return a default value like true so that the UserMenu and DropdownToolbar still work as expected (they use the DrawerNavItem but won't have a Drawer parent to learn the open state from).

Here's some more reading on Context for background: https://reactjs.org/docs/context.html

components/src/core/Drawer/Drawer.tsx
```diff
@@ -282,14 +283,24 @@ const DrawerRenderer: React.ForwardRefRenderFunction<unknown, DrawerComponentPro
...
-            <div className={clsx(defaultClasses.content, classes.content)} style={{ width: contentWidth }}>
-                {getDrawerContents()}
-            </div>
+            <DrawerContext.Provider value={{ isOpen: isDrawerOpen() }}>
+                <div
+                    className={clsx(
+                        defaultClasses.content,
+                        classes.content,
+                        'pxb-drawer-content',
+                        isDrawerOpen() && 'pxb-drawer-open'
```
joebochill, line 297: Can we eliminate these two classes now?

components/src/core/Drawer/DrawerContext.ts
```diff
@@ -0,0 +1,9 @@
+import { createContext, useContext } from 'react';
+
+type DrawerContextType = {
+    isOpen: boolean;
+};
+
+export const DrawerContext = createContext<DrawerContextType>({ isOpen: true });
```
joebochill, line 7: Is there a rationale behind defaulting to `open: true` versus `open: false` or `null`? I don't think it will have an impact since we're only using it internally and we initialize it with a value, but I was just curious.

huayunh: >Is there a rationale behind defaulting to open: true versus open: false or null? 

There is -- the DrawerNavGroup is designed to be used on its own, so we need to provide some initial value to display both the title and subtitle

PR #191, opened 2020-08-28
Publish 4.1.0

joebochill:
Move latest published code into master

-   Fixes RTL support issues in the `<DrawerLayout>`/`<Drawer>`
-   Fixes an issue with hover color alpha channel in `<InfoListItem>`
-   Adds `iconAlign` prop to `<InfoListItem>` to align icon left (default), center or right
-   Adds optional prop to the `<DrawerHeader>` to add a bottom divider
-   Adds `<DropdownToolbar>` component


huayunh:
> When you start using forwardRef in a component library, you should treat it as a breaking change and release a new major version of your library. This is because your library likely has an observably different behavior (such as what refs get assigned to, and what types are exported), and this can break apps and other libraries that depend on the old behavior. 

— https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers

joebochill:
@huayunh 

If you've noticed a different behavior or public API somewhere, please let me know. As far as I can tell, I think we can get by with calling this a new feature (without breaking changes)...
- All of our components were previously functional components, so there is no need to worry about class refs turning into DOM refs
- We didn't have any refs before, so there shouldn't be any concern with refs moving from one DOM element to another
- Our peerDependency for React is already > 16.3
- Our components are all presentational, with no real need for anyone to be using refs in the first place
    - Refs are primarily used to get access to `<inputs>` and media playback elements, etc.

https://stackoverflow.com/questions/53447432/is-adding-forwardref-to-a-function-component-a-breaking-change

PR #194, opened 2020-09-04
Fix icons that should not be flipped

huayunh:
- Fix the icons that should not be flipped. Refer to https://github.com/pxblue/react-design-patterns/pull/30

#### Screenshots:
image
(only the log out icon is flipped, as it carries some directionality)

### DIFF
--- a/demos/storybook/stories/drawer/with-full-config.tsx
+++ b/demos/storybook/stories/drawer/with-full-config.tsx
@@ -28,6 +28,7 @@ import {
     DrawerNavGroupProps,
     NavItem,
 } from '@pxblue/react-components';
+import { getDirection } from '@pxblue/storybook-rtl-addon';
 import { boolean, color, number, select, text } from '@storybook/addon-knobs';
 import { StoryFnReactReturnType } from '@storybook/react/dist/client/preview/types';
 import React from 'react';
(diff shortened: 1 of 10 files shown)


demos/storybook/stories/drawer/with-full-config.tsx
```diff
@@ -218,7 +221,7 @@ export const withFullConfig = (context: DrawerStoryContext): StoryFnReactReturnT
             onClick: (): void => {
                 context.store.set({ selected: schedule });
             },
-            icon: <AirportShuttle />,
+            icon: <AirportShuttle style={getLeftToRightIconDirection()} />,
```
joebochill, line 224: This syntax reads cleaner...do you want to also use it for the 3 stories above for consistency?

demos/storybook/stories/user-menu/within-a-toolbar.tsx
```diff
@@ -16,7 +17,7 @@ export const withinToolbar = (): StoryFnReactReturnType => {
         title: {
             color: Colors.gray[600],
             fontSize: 12,
-            textAlign: 'right',
+            textAlign: getDirection() === 'rtl' ? 'left' : 'right',
```
joebochill, line 20: Is this necessary? Shouldn't the ILI be handling RTL correctly already?

huayunh: Yes, it is necessary. This gets rendered as an inline style. And we want to have a right aligned row

demos/storybook/stories/drawer/with-full-config.tsx
huayunh: As long as it doesn’t hurt copy-ability of the code

demos/storybook/stories/user-menu/within-a-toolbar.tsx
joebochill: Oh I see now, it's reversed (we want right aligned for LTR language)...got it.

joebochill (approved):

Issue #199, opened 2020-10-01
Set id or class on Drawer Items


joebochill:
#### Describe the desired behavior
Items in the drawer should do prop spreading so that users can set things like ids or classes on individual drawer (or userMenu) items.

#### Describe the current behavior
Only the properties defined on NavItem are supported:
![image](https://user-images.githubusercontent.com/29152776/94801702-6d703980-03b4-11eb-9974-2ba1c587286b.png)

#### Is this request related to a current issue?
This is a feature request from the SeedUI project. They want IDs on their drawer and user menu items so that they can find them for their automated tests.

#### Additional Context

We should add `HTMLAttributes<HTMLDivElement>` to the type definition for NavItem. 
~~We might also want to add an `InfoListItemProps` prop to spread props directly to the InfoListItem element as well.~~
Edit: we already have this and it might be the best way to address this issue.

joebochill:
This is a little trickier that just spreading the props to the div.

The `PXBlueDrawerInheritableProperties` cause issues with the destructuring of the props. Once we pull out the DrawerNavItemProps, we're left with a combination of the inheritable props and the div props. If we spread the inheritable props to the div, you'll get console errors for unrecognized props. 

I haven't found a convenient way to programmatically remove the inheritable props from the object (without manually specifying each one...which is not ideal for future maintainability).

If we cant figure out a good way to do that, we could just spread the most popular div props instead of all of them (id, className, style, other?).

joebochill:
InfoListItemProps covers this, see #200

huayunh:
Maybe we gotta have more examples in the readme to cover how this works :P

PR #200, opened 2020-10-01
Feature/spread drawer item

joebochill:
Partially deals with #199.

- Allow user to supply id, style, and className properties to NavItems (so they can be targeted by automated tests)

> Testable via @pxblue/react-components@4.2.0-beta.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## TBD
+
+### Added
+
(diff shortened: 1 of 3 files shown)


joebochill:
I confirmed that the InfoListItem props works well for this. It puts the id on the InfoListItem instead of the parent wrapper div. I have a question out to Jack to see if that will suffice for his tests (I suspect it will), in which case I will close this PR.

joebochill:
Jack confirmed the ILIProps will work for his purposes.

PR #202, opened 2020-10-20
Update default LIT styles

joebochill:
For PXBLUE-1564 .

- Remove the default Overline typography variant for LIT
- If variant is not specified, default to our recommended styles
- If variant is specified, use those styles (plus some for padding and roundness)

The end result should look almost exactly the same unless a user has been setting an explicit variant.

#### Screenshots:
image

### DIFF
--- a/components/src/core/ListItemTag/ListItemTag.tsx
+++ b/components/src/core/ListItemTag/ListItemTag.tsx
@@ -21,7 +21,7 @@ const useStyles = makeStyles((theme: Theme) =>
             borderRadius: theme.spacing(0.25),
             padding: 0,
             paddingLeft: theme.spacing(0.5),
-            paddingRight: theme.spacing(0.5),
+            paddingRight: theme.spacing(0.5) - 1, // to account for extra pixel from letter-spacing
             overflow: 'hidden',
             backgroundColor: (props: ListItemTagProps): string => props.backgroundColor || theme.palette.primary.main,
             color: (props: ListItemTagProps): string => props.fontColor || theme.palette.primary.contrastText,
@@ -29,7 +29,7 @@ const useStyles = makeStyles((theme: Theme) =>
             display: 'inline-block',
         },
         noVariant: {
-            fontWeight: 'bold',
+            fontWeight: 700, // bold
             letterSpacing: 1,
             fontSize: 10,
             lineHeight: '16px',
--- a/components/src/core/ListItemTag/ListItemTag.tsx
+++ b/components/src/core/ListItemTag/ListItemTag.tsx
@@ -15,6 +15,18 @@ export type ListItemTagProps = TypographyProps & {
     label: string;
 };
 
+// See https://material-ui.com/guides/right-to-left/#opting-out-of-rtl-transformation
+declare module '@material-ui/core/styles/withStyles' {
+    // Augment the BaseCSSProperties so that we can control jss-rtl
+    // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+    interface BaseCSSProperties {
+        /**
+         * Used to control if the rule-set should be affected by rtl transformation
+         */
+        flip?: boolean;
+    }
+}
+
 const useStyles = makeStyles((theme: Theme) =>
     createStyles({
         root: {
@@ -27,6 +39,7 @@ const useStyles = makeStyles((theme: Theme) =>
             color: (props: ListItemTagProps): string => props.fontColor || theme.palette.primary.contrastText,
             cursor: (props: ListItemTagProps): string => (props.onClick ? 'pointer' : 'inherit'),
             display: 'inline-block',
+            flip: false, // letter-spacing doesn't flip for RTL, so neither shall our padding hack to offset it
         },
         noVariant: {
             fontWeight: 700, // bold
--- a/components/src/core/ListItemTag/ListItemTag.tsx
+++ b/components/src/core/ListItemTag/ListItemTag.tsx
@@ -1,10 +1,10 @@
 import { Typography, TypographyProps } from '@material-ui/core';
-import { createStyles, makeStyles, Theme, TypographyVariant } from '@material-ui/core/styles';
+import { createStyles, makeStyles, Theme } from '@material-ui/core/styles';
 import React from 'react';
 import PropTypes from 'prop-types';
 import clsx from 'clsx';
 
-export type ListItemTagProps = {
+export type ListItemTagProps = TypographyProps & {
     /* Color of the label background. Default is theme.palette.primary.main */
     backgroundColor?: string;
 
@@ -13,10 +13,7 @@ export type ListItemTagProps = {
 
     /* The string label of the tag. */
     label: string;
-
-    /* The typography variant - we override this to allow for undefined (default styles). */
-    variant?: TypographyVariant;
-} & Omit<TypographyProps, 'variant'>;
+};
 
 const useStyles = makeStyles((theme: Theme) =>
     createStyles({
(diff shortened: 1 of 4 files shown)


joebochill:
So I know I was the one pushing for keeping the variant styles in the react implementation, but they are actually kind of a pain in the butt to deal with. Adding a 'default' option complicates the types and the documentation and I wonder if people are even going to use these variants at all (I'm thinking probably not). I would be okay with having a single implementation of this component and taking out the variant prop.

Thoughts?

joebochill:
Or we could leave the prop in (continue to spread all props to Typography component), but leave our default styles in and make the user override them if they actually want a different variant to look good.

huayunh:
* [ ] Fix Storybook

* [ ] For the default style we might want `height: 16px; display: 'inline-block'` in addition to the listed styles (height of the tag should be 16px according to the design files)

* [ ] The tag is not properly centered in the info list item. We might want to flex it.
Screen Shot 2020-10-20 at 4 11 34 PM

### For the variant logic
I am happy with the way it is implemented. Will it be more consistent to introduce a "default" class for the `classes` props? So the user can do 
```ts
{
    root: blah,
    default: blah, 
    overline: blah,
    body1: blah,
    ...
}
```

I don't get what these lines are doing
image
Isn't the `variant` prop in MUI's `TypographyProps` already `variant?: Variant | 'inherit';`? You are leaving out the 'inherit' option here.

joebochill:
@huayunh 

You've started to unravel the reasons why this is confusing. Introducing the 'default' variant causes issues like:

- What do we call it? Do we actually call it `variant={'default'}` or do we just do these styles if they leave it off entirely?
    - If the former, we have to extend the type definition from MUI for variant (and the types that they export don't exactly match what is in their docs — and they are dropping some of these in their next release). So then we have to keep up with changes in MUI to make sure we are in sync.
        - And then how do we document this? The type of variant is `all of the Typography variants | 'default'`? Or do we list them all out (maintenance headache)? And now this variant property is different from the variant used on other Typography elements (confusion).
    - If the latter, how do you show it in the full config storybook story? We can't call it 'default' in the dropdown because it's not. We could leave it blank, but then how do users know what that means? Or I suppose we could call it 'No variant'. It all seems very convoluted.

joebochill:
My main question is: 
Is there any value in supporting different variants for the ListItemTag?

So we ever anticipate using different styles for these? Keeping in mind that React Native and Angular both have a single style available for these.

joebochill:
> I don't get what these lines are doing
> image
> Isn't the `variant` prop in MUI's `TypographyProps` already `variant?: Variant | 'inherit';`? You are leaving out the 'inherit' option here.

`variant` is a required prop in TypographyProps, so we have to omit it in order to make it optional. Some of the values listed in the documentation are not actually in the TypographyVariant type, so it might need to be something like:

```
variant?: TypographyVariant | 'inherit'
```

huayunh:
>   * If the latter, how do you show it in the full config storybook story? We can't call it 'default' in the dropdown because it's not. We could leave it blank, but then how do users know what that means? Or I suppose we could call it 'No variant'. It all seems very convoluted.

image
https://pxblue-components.github.io/react-dev/?path=/story/components-drawer--with-full-config

_Yes, I prefer the latter._

huayunh:
> `variant` is a required prop in TypographyProps, so we have to omit it in order to make it optional. 

I don't think `variant` is required in [`<Typography />`](https://material-ui.com/api/typography/). I believe all its props are optional. 

image

joebochill:
> I don't think `variant` is required in [`<Typography />`](https://material-ui.com/api/typography/). I believe all its props are optional.

Ah, interesting. I guess it has a default value but is not required. Good to know.

components/src/core/ListItemTag/ListItemTag.tsx
```diff
@@ -40,6 +45,7 @@ const ListItemTagRender: React.ForwardRefRenderFunction<unknown, ListItemTagProp
     const {
         classes: userClasses,
         label,
+        variant,
         // ignore unused vars so that we can do prop transferring to the root element
         /* eslint-disable @typescript-eslint/no-unused-vars */
         fontColor,
```
emclaug2, line 51: These unused props don't need to be deconstructed here; I think you can delete `fontColor` and `backgroundColor`.

emclaug2:
> My main question is:
> Is there any value in supporting different variants for the ListItemTag?
> 
> Do we ever anticipate using different styles for these? Keeping in mind that React Native and Angular both have only a single style available for these.

I like how this component extends the `TypographyProps` to handle variant; we shouldn't have to explicitly list out all available variants.   I've added just one comment, otherwise this looks good to me.

components/src/core/ListItemTag/ListItemTag.tsx
joebochill: We destructure them so that we can pass all of the `otherTypographyProps` to the Typography component. If we don't destructure them, then when we pass `otherTypographyProps` to the Typography component, it will complain that 'backgroundColor is not a known property of Typography'.

The comment here tries to explain that. We do this in a couple of our components. As of yet, I haven't thought of a better way of doing it.

emclaug2: I just ran into this issue while doing some Drawer refactors; this makes sense now.  It's not intuitive though for first-timers; I think there's gotta be a better way to do this, but I cannot think of a solution either.    This is fine for now :D

components/src/core/ListItemTag/ListItemTag.tsx
```diff
@@ -13,22 +13,27 @@ export type ListItemTagProps = {
...
             paddingRight: theme.spacing(0.5),
-            lineHeight: 'inherit',
             overflow: 'hidden',
             backgroundColor: (props: ListItemTagProps): string => props.backgroundColor || theme.palette.primary.main,
             color: (props: ListItemTagProps): string => props.fontColor || theme.palette.primary.contrastText,
             cursor: (props: ListItemTagProps): string => (props.onClick ? 'pointer' : 'inherit'),
+            display: 'inline-block',
+        },
+        noVariant: {
+            fontWeight: 'bold',
```
huayunh, line 32: I am OK either way, but I recall you prefer fontWeight to be 700 instead of bold?

huayunh (changes requested):
## 1
http://localhost:6006/?path=/story/components-list-item-tag--in-info-list-item has strange spacing now, might have to fix that flexing 

image


## 2
Do we want to do the same padding trick to get around with the extra 1px on the right?

## 3
`yarn update:submodule`?

huayunh:
Internationalization? 

I didn't realize that the letter-spacing is not influenced by this, that the spacing is still added to the right of the letters. Or perhaps when they are on an actual right-to-left language, the letter spacing will get added to the left of the characters automatically?

image
image

emclaug2 (approved):

joebochill:
> Internationalization?
> 
> I didn't realize that the letter-spacing is not influenced by this, that the spacing is still added to the right of the letters. Or perhaps when they are on an actual right-to-left language, the letter spacing will get added to the left of the characters automatically?

Just did a quick check on this and letter-spacing adds the space on the right, even for RTL languages. So the padding hack should be fine for either language direction. I'll quickly update the storybook margin to work for RTL.

huayunh:
What a strange API design... shows no love for the Hebrew world. 


_Against stupidity the gods themselves contend in vain_

joebochill:
> What a strange API design... shows no love for the Hebrew world.
> 
> _Against stupidity the gods themselves contend in vain_

I read up a bit while I was implementing these changes and apparently the letter-spacing behavior in CSS is basically completely nonsensical according to the internet. It's been a topic of complaint since the beginning of time.

huayunh (approved):

huayunh:
> Just did a quick check on this and letter-spacing adds the space on the right, even for RTL languages. 

So should the spacing hack change when the theme direction change? 

Right now it is:
Theme: LTR
- padding left: 4
- padding right: 3

Theme RTL
- padding left: 3
- padding right: 4

But since the padding is always applied to the left, should we enforce the rule to always be 4-3?

PR #206, opened 2020-12-01
Hero margin 5px -> 8px

huayunh:
## Changes proposed in this Pull Request:
- Replaced a magic number to use `theme.spacing(1)` to match design files

## Screenshots
| Present (9px) | Proposed (12px) |
| -- | -- |
|image | image |

_I know... this is just some designer itch_

## Additional Comment

I was looking into RN too, but it seems that RN is using some different pixels / dimensions than ng / React, so I didn't touch RNCL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,11 @@
 ## TBD
 
 ### Changed
--   Updated styles for default `<ListItemTag>`
+
+-   Updated styles for default `<ListItemTag>` and `<Hero>`
 
 ### Added
+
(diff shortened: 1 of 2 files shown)


joebochill:
12px looks like a pretty big gap. Would we not prefer to go 9px -> 8px instead?

huayunh:
All the design files are currently using ~12px. All = WAS, IPPE, GP and DSM. 
(To be precise, they are doing 11px, and I don't know why)

joebochill (approved):

PR #207, opened 2020-12-01
Fix mismatched drawer transitions

### LINKED ISSUE #201: Persistent Drawer's transition animation is "detached"
### Describe the bug

Drawer's transition animation seems detached from it's layout; it is likely that we used two different css transition style.

### What is the expected behavior?

Drawer's expand/collapse should match perfectly

### What are the steps to reproduce?
1. Spin up the react showcase demo
2. Collapse / expand the drawer by either hover on drawer items or by clicking on the menu icon

### Screenshots or links to minimum reproduction example

Video set to x0.5 speed to illustrate the problem

drawer-transition

### Environment

MacOS Chrome

joebochill:
Helps with #201.

- Update the DrawerLayout transition to match the timing of the Drawer transition

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,15 @@
 # Changelog
 
-## TBD
+## TBD (Not published yet)
 
 ### Changed
 -   Updated styles for default `<ListItemTag>`
 
 ### Added
 -   Adds `hidden` prop to the Drawer `NavItem` to conditionally omit items from the `<Drawer>` or `<UserMenu>`.
+
+### Fixed
+-   Mismatched transition speeds on the `<Drawer>` and `<DrawerLayout>`
 
 ## v4.1.1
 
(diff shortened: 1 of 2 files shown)


joebochill:
This seems to help with the transition, but if you get to clicking around on the drawer a bunch, you can still see the 'gap' occasionally. The transitions are the same, but I suspect there is a race condition when the variables get set to trigger them (drawer gets open/closed state set right away through the prop change, but the DrawerLayout has to wait until the Context is updated). We might need to explore moving some of the drawer logic into the context provider so that these get triggered at the exact same time.

huayunh (approved):
Code looks good. Showcase spins good yet still with some transition animation delay — probably just a matter of my old machine at this point.

Do we want to use these transitions? I am happy with the way it is right now. 
image

joebochill:
> Code looks good. Showcase spins good yet still with some transition animation delay — probably just a matter of my old machine at this point.
> 
> Do we want to use these transitions? I am happy with the way it is right now.

It's probably not a bad idea to use these...it's a little more involved though...we'd have to change the transition based on whether the drawer is open or closed to get the different timings. Right now we just set the transition once and use the same timing for entering and exiting. I'll log an enhancement for this.

PR #208, opened 2020-12-02
Update icon import syntax to efficient version

joebochill:
- Update icon import syntax to reduce bundle size

Available for testing as @pxblue/react-components@4.2.0-beta.1


huayunh (approved):
Code looks all good

jeffvg (approved):
verified with beta package and build is half the size

PR #213, opened 2020-12-28
Update images; use flex box to align icon buttons

### LINKED ISSUE #180: Check for IconButton vs. Icon in Dropdown Toolbar
User should be able to pass in an IconButton or Icon to the navigationIcon prop and have the toolbar styled correctly without additional styles required.

If the user passes in an IconButton, they will need to add negative margins to it for the icon to properly align to the keylines in the toolbar.

#### Additional Context
We should be able to do some sort of check of the displayType or displayName to determine if we're dealing with an IconButton, Icon, or something else and then add the negative margins automatically for the IconButton.

huayunh:
Fixes #180 ; shall be merged with https://github.com/pxblue/react-showcase-demo/pull/7

- Making sure both plain icons and iconButtons are flexed to the center
- Added a negative margin between title and subtitle to match with the angular implementation
- Reset the margins of ListItemText so that the total height won't make the app bar overflow to 60px on mobile
- Updated the images in the dropdown toolbar readme

#### Screenshots:
The `edge` trick mentioned in the discussion works... but only for the left margin of the icon. The right margin of the icon is not getting that -12px margin when you set edge=start, and it looks like this
image
(The bar on the top comes from DSM)

So I used a flexbox to center whatever the user passes in. 

The current solution does not welcome the user to set edge = start, as this will add that negative margin and nudge the icon to the left, like this:
image
I don't know if this is the ideal solution...


CHANGELOG.md
```diff
@@ -2,9 +2,13 @@
 
 ## 4.2.1 (not published)
 
+### Changed
+
+-   Updated styles for default `<DropdownToolbar>`
```
joebochill, line 7:
```suggestion
-   Updated styles for `<DropdownToolbar>` to support Icon Button or plain icon for `navigationIcon`.
```

joebochill (approved):
I think this is a nice clever solution that's much simpler and more flexible than what I was suggesting in the issue discussion (trying to read in the display name, etc.). Nice job.

I think the edge property thing is fine. Our implementation actually makes it easier for the user...they don't have to know anything about the edge property at all because it just works without it.

PR #214, opened 2020-12-28
Use theme transition duration for drawer

### LINKED ISSUE #210: Use recommended animation speeds for drawer
Use the recommended transition speeds for opening and closing the drawer (entering and exiting).

Using the 'standard' speed for both in and out.

https://github.com/pxblue/react-component-library/pull/207#pullrequestreview-544130015

huayunh:
Fixes #210

- Updated drawer layout and drawer to use `theme.transitions.duration.enteringScreen` and `leavingScreen`


components/src/core/DrawerLayout/DrawerLayout.tsx
```diff
@@ -59,7 +65,10 @@ const DrawerLayoutRender: React.ForwardRefRenderFunction<unknown, DrawerLayoutPr
         >
             <div ref={ref} className={clsx(defaultClasses.root, classes.root)} {...otherDivProps}>
                 <div className={clsx(defaultClasses.drawer, classes.drawer)}>{drawer}</div>
-                <div className={clsx(defaultClasses.content, classes.content)} style={style}>
+                <div
+                    className={clsx(defaultClasses.content, classes.content, padding === 0 && defaultClasses.expanded)}
```
joebochill, line 69: for conditional classes in clsx, we usually prefer the object syntax:
```suggestion
                    className={clsx(defaultClasses.content, classes.content, {[defaultClasses.expanded]: padding === 0})}
```

components/src/core/Drawer/Drawer.tsx
```diff
@@ -8,18 +8,30 @@ import { InfoListItemProps as BaseInfoListItemProps } from '../InfoListItem';
...
+        root: {
+            transition: theme.transitions.create('width', { duration: theme.transitions.duration.leavingScreen }),
+            minHeight: '100%',
+            '&$expanded': {
+                transition: theme.transitions.create('width', {
+                    duration: theme.transitions.duration.enteringScreen,
+                }),
+            },
+        },
+        expanded: {},
```
joebochill, line 22: New classes need to be added to the readme docs

components/src/core/DrawerLayout/DrawerLayout.tsx
joebochill: applies to the one above too

joebochill (changes requested):

components/src/core/Drawer/Drawer.tsx
huayunh: Do we want to add “expanded” as part of a user class too? (Not just “defaultClasses?)

joebochill: I would say yes, if it's a class that gets applied, we should give them a way to customize it if desired.

huayunh:
I just realized that DrawerLayout doesn't really have access to whether a drawer is opened or not; when the variant is set to 'persistent', padding===0 never means "drawer is collapsed". 

Perhaps we should just close this or wire the drawer's open state into the drawer layout context provider

joebochill:
> I just realized that DrawerLayout doesn't really have access to whether a drawer is opened or not; when the variant is set to 'persistent', padding===0 never means "drawer is collapsed".
> 
> Perhaps we should just close this or wire the drawer's open state into the drawer layout context provider

Hmm good call. You're right padding=0 is not a good check for the persistent drawer. When I was messing around with the out of sync animations I implemented a pretty quick and dirty version of a context provider that shared the open state with the Layout, but it was kinda messy and I'm not sure if that's the best way to do it. We may need to have a little more discussion around it to decide on a better approach. 

You may want to add a little more context to #210 to explain the issue. 

Does this break with the persistent drawer variant (e.g., animations using different timings) or is it something that's not very noticeable?

huayunh:
It is insignificant — I didn't even notice the classes were missing until I looked into the inspector.

joebochill:
> It is insignificant — I didn't even notice the classes were missing until I looked into the inspector.



> It is insignificant — I didn't even notice the classes were missing until I looked into the inspector.

Then I think we can push this update in (after readme update) and make a new issue/enhancement to get the open/closed state up into the drawer layout via context (or we can hijack the existing context to pass the required padding and the open/closed state).

joebochill (approved):

PR #219, opened 2021-01-12
Add publish step for master merges

emclaug2:
- Npm publish on master merges

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,6 +67,10 @@ jobs:
                   command: |
                       cd demos/storybook
                       yarn install --frozen-lockfile
+
+                      # Install the packed version of @pxblue/react-components
+                      PACKAGED_VERSION=`node -p "require('../../dist/package.json').version"`
+                      yarn add ../../dist/pxblue-react-components-$PACKAGED_VERSION.tgz
             - save_cache:
                   name: Save Cache
                   paths:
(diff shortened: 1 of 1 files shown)


joebochill (commented):
So this one actually introduces a new challenge. The deployed version of storybook should always be using the latest published version of the package from NPM. To this end, on line 90, we actually override all of the local code with the version from NPM (so the current implementation is always going to deploy storybook with the (N-1)th version).

I think we may need to do some reordering of the build steps/jobs.

For the master branch, we should 

1. verify that storybook builds with the local code
2. then publish the new version to NPM
3. then build and deploy the storybook with the version from NPM

huayunh:
will there be a racing condition if we pull the version immediately from NPM after we publish a new package?

emclaug2:
> will there be a racing condition if we pull the version immediately from NPM after we publish a new package?

Shouldn't be.  We'll run these as separate jobs that'll wait for the publish step to complete before starting a build.  I'm still thinking about the best way to fix this without slowing down our ci builds too much.

emclaug2:
@joebochill what are your thoughts on this? 

I wanted to minimize the amount of refactoring, so our order of jobs is still:

```
build
| ---> showcase 
|                         \
|                           \
| ---> storybook ---> publish
                          \
                             \--> deploy

```

This changes the install step to pack and unpack the package locally: 
```
 if [ $CIRCLE_BRANCH == $MASTER_BRANCH ]; then
    cd ../../dist
    MASTER_VERSION=`node -p "require('./package.json').version"`
    npm pack
    cd ../demos/storybook
    npm install ../../dist/pxblue-react-components-$MASTER_VERSION.tgz --production
  fi
```

joebochill:
> @joebochill what are your thoughts on this?
> 
> I wanted to minimize the amount of refactoring; this changes the install step to pack and unpack the package locally:
> 
> ```
> # Install the packed version of @pxblue/react-components
> PACKAGED_VERSION=`node -p "require('../../dist/package.json').version"`
> yarn add ../../dist/pxblue-react-components-$PACKAGED_VERSION.tgz 
> ```
> 
> This is an alternative to installing the package through npm after the package is deployed.

I think this is an excellent idea. In theory the packed version should install exactly the same way as if it were from NPM and as you mentioned it will be a good way to do it for all branches (eliminating the awkward copy step, which doesn't correctly pull in other dependencies, etc.). Nice find.

joebochill (approved):

PR #220, opened 2021-01-13
Feature/open on hover

emclaug2:
- Add `openOnHover` prop to Drawer

https://pxblue-react-library.web.app/?path=/story/components-drawer--with-full-config

### DIFF
--- a/components/src/core/Drawer/Drawer.tsx
+++ b/components/src/core/Drawer/Drawer.tsx
@@ -142,6 +142,9 @@ export type DrawerComponentProps = {
     // Controls the open/closed state of the drawer
     open: boolean;
 
+    // Enables Drawer to automatically open on hover for persistent variants.
+    openOnHover?: boolean;
+
     // Toggles the drawer side border instead of a drop shadow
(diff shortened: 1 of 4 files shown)


components/src/core/Drawer/Drawer.tsx
```diff
@@ -284,11 +288,15 @@ const DrawerRenderer: React.ForwardRefRenderFunction<unknown, DrawerComponentPro
                 <div
                     style={{ flexDirection: 'column', flex: '1 1 0px', display: 'flex' }}
                     onMouseEnter={(): void => {
```
joebochill, line 290: Looks good, I would just change one thing...put the conditionals outside instead of inside the events. So we don't have a bunch of events firing off that aren't doing anything, i.e.:

```
onMouseEnter={ onOpenHover ? (): void => { hoverDelay = setTimeout(...) } : undefined }
```

emclaug2: Change should be pushed out now.

joebochill (approved):

PR #222, opened 2021-01-14
Add Rail variant for Drawer

joebochill:
For PXBLUE-1591

- Add a rail variant for the drawer
- Includes a new DrawerRailItem component
- Update docs and demos

#### Screenshots:
image


components/src/core/Drawer/Drawer.tsx
```diff
@@ -145,10 +151,13 @@ export type DrawerComponentProps = {
...
     sideBorder?: boolean;
 
+    // Drawer variant type
+    variant?: 'persistent' | 'permanent' | 'temporary' | 'rail';
+
     // Sets the width of the drawer (in px) when open
     width?: number;
 } & PXBlueDrawerNavGroupInheritableProperties &
-    Omit<DrawerProps, 'translate'>;
+    Omit<Omit<DrawerProps, 'translate'>, 'variant'>;
```
emclaug2, line 160: Does this work?

`Omit<DrawerProps, 'translate' | 'variant'>` ?

joebochill: Negative. The `Omit` helper type only allows you to omit a single property.

joebochill: `Pick` let's you specify multiple in that way, but not `Omit` for some reason.

emclaug2 (commented):
Looks like the ripple effect is missing whenever select a rail item.

components/src/core/Drawer/Drawer.tsx
emclaug2: Ah sounds good!

emclaug2:
This looks really nice so far; I like that you made a separate item for the DrawerRail items.

components/src/core/Drawer/Drawer.tsx
joebochill: Actually, it's looking like your suggestion may be possible now...looking into it.

emclaug2:
Screen Shot 2021-01-14 at 2 39 37 PM

Little funky behavior on the dark theme.
RTL Drawer header is also offset from center when rail.

components/src/core/Drawer/Drawer.tsx
joebochill: It does seem to work...I wonder if that's something they fixed in typescript 4. 

We can fix this in doc-it now:

image

joebochill:
> Little funky behavior on the dark theme.
> RTL Drawer header is also offset from center when rail.

Weird...doesn't look like that in the showcase. I'll have a look.
image

joebochill:
It has something to do with the shared nav item lists between stories....depending on the order that you navigate, sometimes it has that weird behavior and sometimes not.

emclaug2:
> It has something to do with the shared nav item lists between stories....depending on the order that you navigate, sometimes it has that weird behavior and sometimes not.

I think it'd be best to make a separate issue for this since it's not an issue within the component.

jeffvg:
will divider be added in for this?
Wondering.. will the hover state like the angular one be part of this when condensed?

image

joebochill:
> > It has something to do with the shared nav item lists between stories....depending on the order that you navigate, sometimes it has that weird behavior and sometimes not.
> 
> I think it'd be best to make a separate issue for this since it's not an issue within the component.

I fixed it. There was a dependency missing from one of the callback hooks, so when navigating around, the styles weren't getting recalculated.

joebochill:
> will divider be added in for this?
> Wondering.. will the hover state like the angular one be part of this when condensed?

Done

huayunh:
`git push`?

components/src/core/Drawer/DrawerRailItem.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+    const combine = useCallback(
+        (className: keyof DrawerRailItemClasses): string => clsx(defaultClasses[className], classes[className]),
+        [defaultClasses, classes]
+    );
+
+    const getIcon = useCallback((): JSX.Element | undefined => {
+        if (icon) {
+            return <Avatar className={combine('icon')}>{icon}</Avatar>;
+        }
```
huayunh, line 199: Somehow when I try to wrap icons in a Badge, the badge doesn't show up normally. I suppose we are not going to use an Avatared icon for navigation rail anyway, so maybe we should abandon this avatar logic? 

image

This corresponds to the code 

``` tsx
                        {
                            title: titleList[0],
                            itemID: titleList[0],
                            icon: <Badge color={'primary'} badgeContent={'2'}><ListIcon className={clsx({ [classes.iconFlip]: rtl })} /></Badge>,
                            onClick: (): void => {
                                navigate(0);
                                if (xsDown) dispatch({ type: CLOSE_DRAWER });
                            },
                            statusColor: 'red',
                        },
```

in showcase

components/src/core/Drawer/DrawerRailItem.tsx
```diff
@@ -0,0 +1,259 @@
...
+    activeItem?: string;
+
+    // toggles the condensed style
+    condensed?: boolean;
+
+    // sets whether to hide the nav item
+    hidden?: boolean;
+
+    // icon on the left
+    icon?: JSX.Element;
```
huayunh, line 40: `icon` should be mandatory in a rail

demos/storybook/stories/drawer/with-different-variants.tsx
```diff
@@ -9,7 +9,9 @@ import { Menu } from '@material-ui/icons';
 export const withDifferentVariants = (context: DrawerStoryContext): StoryFnReactReturnType => (
     <Drawer
         open={boolean('open', true)}
-        variant={select('variant', ['permanent', 'persistent', 'temporary'], 'permanent')}
+        variant={select('variant', ['permanent', 'persistent', 'temporary', 'rail'], 'permanent')}
```
huayunh, line 12: What about the full config and the within drawer layout story?

demos/storybook/stories/drawer/with-different-variants.tsx
```diff
@@ -9,7 +9,9 @@ import { Menu } from '@material-ui/icons';
...
+        condensed={boolean('condensed (rail only)', false)}
+        divider={boolean('divider', false)}
```
huayunh, line 14: why `divider` in a story about variants?

huayunh (changes requested):
The style looks good

May want to note in the drawer.md that certain Shared Props do not apply to the railed items

components/src/core/Drawer/DrawerRailItem.tsx
huayunh: And should we make title optional for rails?

components/src/core/Drawer/DrawerRailItem.tsx
```diff
@@ -0,0 +1,259 @@
...
+        [onItemSelect, onClick]
+    );
+
+    const innerContent = (
+        <ButtonBase
+            ref={ref}
+            {...buttonBaseProps}
+            className={clsx(defaultClasses.root, classes.root, {
+                [defaultClasses.condensed]: condensed,
+                [classes.condensed]: condensed,
```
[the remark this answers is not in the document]
joebochill: This appears to be a larger issue. Any time a class is optional (in the `classes` prop) and it is not provided, it's putting undefined. You can see this happening in the `activeTitle` class too on the title.

I can see about addressing this for the DrawerRailItem, but we should log another issue to fix this in the other components.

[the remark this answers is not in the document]
joebochill: it doesn't happen when you use clsx standard syntax (falseys are discarded), but when you use the object syntax to do things conditionally, we need an extra check to see if the class is defined (instead of undefined).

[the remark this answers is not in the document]
huayunh: so can we do `[classes.condensed]: condensed & classes?.condensed`?

components/src/core/Drawer/DrawerRailItem.tsx
```diff
@@ -0,0 +1,259 @@
...
+    return createStyles({
+        root: {
+            width: RAIL_WIDTH,
+            minHeight: RAIL_WIDTH,
+            position: 'relative',
+            display: 'flex',
+            flexDirection: 'column',
+            alignItems: 'center',
+            justifyContent: 'center',
+            padding: `${theme.spacing(2)}px ${theme.spacing(1)}px`,
```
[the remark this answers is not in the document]
joebochill: Yeah, the status stripe is 6px, so if we give them an additional 4px of padding, they could overlap (label will go behind the stripe) by 2px. So I would say not to add the additional padding, unless we rethink how they should look with a status.

components/src/core/Drawer/DrawerRailItem.tsx
joebochill: The Avatar is there to maintain a consistent sizing. I can turn on overflow so you can see a badge that overextends the size.

demos/storybook/stories/drawer/with-different-variants.tsx
joebochill: Jeff specifically asked for this to match the Angular one

components/src/core/Drawer/DrawerRailItem.tsx
huayunh: That works too. The badges should work out of the box

demos/storybook/stories/drawer/with-different-variants.tsx
huayunh: What was the rationale behind the Angular one to add a divider field? @jeffvg

joebochill:
@huayunh You have brought up some interesting points that are worth discussing.

As you have noticed, there are a number of properties that the RailItem does not support and you have proposed a number of new properties (e.g., enableTooltipOnCondensed) that would only apply the the RailItem. This introduces some challenges in:
- **Documentation**: We have to add a bunch of special cases saying "the drawer supports the following item format, except if you are using the rail in which case these 15 props don't work, and these other 15 props are now required instead of optional, and these other required props are now optional, and there are also these other 15 props, but those are only for the Rail".
    - It gets very confusing quickly
- **Implementation**: I'm assuming that we want it to be convenient to switch between the different variants of the Drawer. If we suddenly start changing the type defs and which props are required or not, then users won't be able to simply swap out the `variant`, but also make a bunch of conditional updates to the item definitions.

With this is mind...is the NavigationRail even a Drawer variant? If 50% of the props are not supported and there are other special considerations, does it make sense to have it as a variant or should it be a completely separate component?

huayunh:
IKR...
Problem is that the rail also share 80% of the logic with the rest of the drawer. 

How about inventing a `<DrawerRailGroup>` component that completely run on its own set of `RailItemProps`, instead of just an internal NavRailItem embedded inside the existing `DrawerNavGroup` ? Will that help with the props sharing?

components/src/core/Drawer/DrawerRailItem.tsx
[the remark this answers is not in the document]
huayunh: Sounds like another docit patch?

components/src/core/Drawer/DrawerRailItem.tsx
```diff
@@ -0,0 +1,259 @@
...
+export type DrawerRailItemProps = Pick<
+    PXBlueDrawerInheritableProperties,
+    | 'activeItemBackgroundColor'
+    | 'activeItemFontColor'
+    | 'activeItemIconColor'
+    | 'divider'
+    | 'itemFontColor'
+    | 'itemIconColor'
+    | 'onItemSelect'
+    | 'ripple'
```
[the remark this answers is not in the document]
joebochill: We can do the same thing with `ButtonBaseProps`...I'll update the docs

components/src/core/Drawer/DrawerRailItem.tsx
joebochill: I can make the type mandatory, but it won't have a real appreciable effect since users don't use the DrawerRailItem directly...all of the props passed to it are controlled by the DrawerNavItem, which currently uses NavItem as the type for items (that's a whole different can of worms...i tried to do NavItem | RailItem....but then all of the differences between the two started coming up everywhere...red squigglies like you wouldn't believe).

I think for the Rail, if they don't want a title, they can pass in `""` or use the condensed option.

components/src/core/Drawer/DrawerRailItem.tsx
```diff
@@ -0,0 +1,259 @@
...
+                    {title}
+                </Typography>
+            )}
+            {/* Divider */}
+            {props.divider && <Divider className={combine('divider')} />}
+        </ButtonBase>
+    );
+
+    return hidden ? null : condensed ? (
+        <Tooltip title="Add" placement="right">
```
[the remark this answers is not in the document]
joebochill: I agree it would be nice to be able to turn it off...trying to think of the best way to do that. I'm hesitant to add a prop that is new (i.e. not supported by the NavItem), because I have to find some way to pass it in that's not through the `items` prop, or do the battle with types we're talking about below.

We could add a `showTooltip` prop to the NavItem type and implement (optional) tooltips for regular drawer items and rail items. Thoughts?

demos/storybook/stories/drawer/with-different-variants.tsx
jeffvg: I just remember when we decided to add it for the Angular rail and to simply match up what we can offer between the two frameworks.

huayunh:
Or maybe we should just expose the drawer item component?
```tsx
<Drawer>
    <DrawerHeader />
    <DrawerSubHeader />
    <DrawerGroupHeader />
    <DrawerItem />
    <DrawerItem />
    <DrawerFooter />
</Drawer>
```

^ This is how ListItem, ListItemText, Accordion, AccordionSummary are structured. So instead of having everything as a prop, we just make another component out of it 👏

joebochill:
> IKR...
> Problem is that the rail also share 80% of the logic with the rest of the drawer.
> 
> How about inventing a `<DrawerRailGroup>` component that completely run on its own set of `RailItemProps`, instead of just an internal NavRailItem embedded inside the existing `DrawerNavGroup` ? Will that help with the props sharing?

I think the easiest thing to do for implementation and documentation is to treat the RailItem as a subset of the NavItem and have a section in the README that says: "RailItem only uses the following properties from the NavItem type:". 

Changing the optional/required props I think gets too unwieldy. I would lean towards making the majority of the props optional and making sure we have sensible defaults. 

Recommendations on the design side can be different from the requirements in the component I think. For example, we can say "When using a navigation rail, all of your items must have an icon, and when using a NavItem all of your items must have a title", but the component could still support items without an icon or without a title.
image

joebochill:
> Or maybe we should just expose the drawer item component?
> 
> ```tsx
> <Drawer>
>     <DrawerHeader />
>     <DrawerSubHeader />
>     <DrawerGroupHeader />
>     <DrawerItem />
>     <DrawerItem />
>     <Spacer />
>     <DrawerFooter />
> </Drawer>
> ```
> 
> ^ This is how ListItem, ListItemText, Accordion, AccordionSummary are structured. So instead of having everything as a prop, we just make another component out of it 👏

I thought about that too.....going in a more declarative structure (like Angular). That forces more of the logic onto the end user though. I think it's worth exploring, but it's probably bigger than a breadbox and something we'll want to look at outside of the scope of this PR. We can put a spike on the JIRA backlog for it. I suspect we could probably support both ways (similarly to how we take in a title prop for the header, but if the user wants to go crazy they can build their own content).

emclaug2 (approved):

components/src/core/Drawer/DrawerRailItem.tsx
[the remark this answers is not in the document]
joebochill: Or we could do a thinner status stripe. Or draw the text on top of the status stripe. Open to suggestions

components/src/core/Drawer/DrawerRailItem.tsx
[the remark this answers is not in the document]
joebochill: I'm inclined to leave the prop out for now (since we currently match Angular) and I can make an enhancement to discuss if we want to give users the ability to enable/disable tooltips and implement it in both frameworks. Is that cool?

jeffvg:
I know we have a list of items in the nav but can we do anything with this?

image

joebochill:
> I know we have a list of items in the nav but can we do anything with this?

I could make the horizontal scrollbar disappear by disabling the overflow, but I wouldn't want to prevent users from scrolling vertically. That's kind of an annoying thing with how windows does scrollbars. How does the angular one look? I'm assuming it's the same?

jeffvg:
> > I know we have a list of items in the nav but can we do anything with this?
> 
> I could make the horizontal scrollbar disappear by disabling the overflow, but I wouldn't want to prevent users from scrolling vertically. That's kind of an annoying thing with how windows does scrollbars. How does the angular one look? I'm assuming it's the same?

The angular story with full config doesn't include the variants for drawer in knobs....

joebochill:
> The angular story with full config doesn't include the variants for drawer in knobs....

What about in the in a drawer layout story if you make your browser window really short (may have to put inspector on the bottom to get small enough since the demo only has two items in it)?

jeffvg:
> > The angular story with full config doesn't include the variants for drawer in knobs....
> 
> What about in the in a drawer layout story if you make your browser window really short (may have to put inspector on the bottom to get small enough since the demo only has two items in it)?

yep, angular rail is the same. nice big chunky scroll bar

components/src/core/Drawer/DrawerRailItem.tsx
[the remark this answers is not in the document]
joebochill: This is sort of as bad as it gets if we go to 4px and leave everything else as is: 
image
Any longer and the text wraps to the next line.

joebochill:
> yep, angular rail is the same. nice big chunky scroll bar

Let's log a new issue for that then (maybe one in JIRA since we'll need to think about it a little bit). Scroll bars are notoriously a pain, but I agree that they look pretty awful and maybe there's something we can do about it.

components/src/core/Drawer/DrawerRailItem.tsx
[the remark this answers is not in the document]
huayunh: `(props) => props.statusColor ? <8px> : <4px>` was the hack I was thinking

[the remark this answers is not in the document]
joebochill: Are you worried about the possibility of a label being forced to wrap when a status color is activated? Like if something goes in and out of alarm the label could switch between wrapping and not wrapping.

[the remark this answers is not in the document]
joebochill: https://user-images.githubusercontent.com/29152776/104783255-9f36c180-5753-11eb-9e1d-4ac1983e702c.mov

components/src/core/Drawer/DrawerRailItem.tsx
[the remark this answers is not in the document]
joebochill: https://github.com/pxblue/react-component-library/issues/223

components/src/core/Drawer/DrawerRailItem.tsx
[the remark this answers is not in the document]
joebochill: I've dropped it to 4px to match the spec and also avoid the text wrapping and unwrapping. If you would prefer the wrapping over the potential 2px overlap with the status stripe, I can swap it back, just let me know.

huayunh (changes requested):
* Rail needs a hover state
* Use this for your rail item picture: 
image
> I've dropped it to 4px to match the spec and also avoid the text wrapping and unwrapping. If you would prefer the wrapping over the potential 2px overlap with the status stripe, I can swap it back, just let me know.

I would personally prefer the conditional padding thing. It is quite rare in the real use for a status stripe to be turned on off on off... like that in the video. 

Other stuff looks good

joebochill:
> * Rail needs a hover state

Angular Rail does not do a hover and there is not one defined in the stickersheet. What do you want to use for it (remember that users can set any background color they want)?

huayunh:
Just keep the consistency with other drawer variant's style, and log an issue for ng if Angular doesn't

docs/Drawer.md
```diff
@@ -380,20 +415,30 @@ import { Link } from 'react-router-dom';
...
     </DrawerBody>
 </Drawer>
 ```
 
+: 
+
+```
+
+```
+
```
huayunh, line 441: Remove those?

joebochill:
> Just keep the consistency with other drawer variant's style, and log an issue for ng if Angular doesn't

Okay, I popped the default gray hover color if the item has an onClick event defined. I may allow for custom background colors as part of the refactor effort, in which case we can steal the logic from the ILI which does a darken effect on whatever the background color is for the hover.

huayunh (approved):

PR #224, opened 2021-01-18
Audit suggestions

emclaug2:

Channel Value Changes:
-  Change full config icon color to red[500].

Drawer Changes:
- Centered footer image in `with footer` story
- Made `with full config` bg color blue[500] instead of gold.
- Some knob name changes

Empty State Changes:
- Made action button outlined variant in both button examples

InfoListItem Changes:
- Font and Bg color changes to be more app-reaslistic.

Scorecard Changes:
- Replace all red[500] bg color headers with blue[500]

UserMenu Changes:
- Remove blue[50] menu bg color from `with full config` story.

### DIFF
--- a/demos/storybook/stories/channel-value/with-full-config.tsx
+++ b/demos/storybook/stories/channel-value/with-full-config.tsx
@@ -10,8 +10,8 @@ import { getLeftToRightIconTransform } from '../../src/utils';
 export const withFullConfig = (): StoryFnReactReturnType => {
     const value = text('value', '123');
     const units = text('units', 'hz');
-    const textColor = color('color', Colors.red[500]);
-    const iconColor = color('icon.htmlColor', Colors.black[500]);
+    const textColor = color('color', Colors.black[500]);
+    const iconColor = color('icon.htmlColor', Colors.red[500]);
     const icon = boolean('Show Icon', true) ? (
         <Trend htmlColor={iconColor} style={getLeftToRightIconTransform()} />
     ) : (
--- a/demos/storybook/stories/drawer/with-basic-config.tsx
+++ b/demos/storybook/stories/drawer/with-basic-config.tsx
@@ -1,4 +1,4 @@
-import { Accessibility, Menu, NotificationsActive, Person, Today, Dashboard, LocalLibrary } from '@material-ui/icons';
+import { Accessibility, Menu, NotificationsActive, Person, Today, Gavel, Settings } from '@material-ui/icons';
 import { Drawer, DrawerBody, DrawerHeader, DrawerNavGroup, NavItem } from '@pxblue/react-components';
 import { boolean, text } from '@storybook/addon-knobs';
 import { StoryFnReactReturnType } from '@storybook/react/dist/client/preview/types';
@@ -31,14 +31,14 @@ export const navGroupItems1: NavItem[] = [
 
 export const navGroupItems2: NavItem[] = [
     {
-        title: 'Overview',
+        title: 'Settings',
         itemID: '2-1',
-        icon: <Dashboard />,
+        icon: <Settings />,
     },
     {
-        title: 'Manuals',
+        title: 'Legal',
         itemID: '2-2',
-        icon: <LocalLibrary />,
+        icon: <Gavel />,
     },
 ];
 
--- a/demos/storybook/stories/drawer/with-custom-header.tsx
+++ b/demos/storybook/stories/drawer/with-custom-header.tsx
@@ -17,14 +17,24 @@ export const withCustomHeader = (context: DrawerStoryContext): StoryFnReactRetur
             backgroundOpacity={0.5}
             icon={<MenuIcon />}
             titleContent={
-                <div style={{ display: 'flex', zIndex: 1, paddingLeft: '20px', paddingTop: '15px' }}>
+                <div
+                    style={{
+                        display: 'flex',
+                        justifyContent: 'space-between',
+                        zIndex: 1,
+                        padding: '0 16px',
+                        alignItems: 'center',
+                        width: '100%',
+                        height: '100%',
+                    }}
+                >
                     <div>
                         <Typography variant="subtitle2">Customizable</Typography>
                         <Typography variant="h6" style={{ marginTop: '-10px' }}>
                             Header Content
                         </Typography>
                     </div>
-                    <ListItemTag style={{ marginLeft: 72 }} label={'v1.0.3'} />
+                    <ListItemTag style={{ marginBottom: 16 }} label={'v1.0.3'} />
                 </div>
             }
         />
--- a/demos/storybook/stories/drawer/with-custom-header.tsx
+++ b/demos/storybook/stories/drawer/with-custom-header.tsx
@@ -6,6 +6,7 @@ import { StoryFnReactReturnType } from '@storybook/react/dist/client/preview/typ
 import React from 'react';
 import { DrawerStoryContext } from './util';
 import { navGroupItems1 } from './with-basic-config';
+import { ListItemTag } from '@pxblue/react-components';
 
 const farmBgImage = require('../../assets/farm.jpg');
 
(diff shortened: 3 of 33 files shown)


emclaug2:
I looked at Kyle's comments but it seems like most recommendations were drawer-related and fixed with the style updates that @joebochill did.   If there are any other things that you'd like to see fixed in storybook, let me know and it can be added to this PR.

emclaug2:
Since I changed the action button type should icon change too?

A (current)
Screen Shot 2021-01-18 at 5 30 07 PM

B
Screen Shot 2021-01-18 at 5 30 17 PM

huayunh:
I'd prefer an uncircled version. And remember to use the `startIcon` prop for a better spacing

huayunh:
image

The 3 dot icon is usually the right-most icon.

huayunh:
image
note the font weight difference in the body list

huayunh:
pick something less outrageous for the font color... gray or black 500
image

huayunh:
And instead of "Account", say "Settings" for the menu item

joebochill:
Overall looking good. I did a quick run through of all of the stories and noticed just a few things that stood out to me:

- Drawer 
    - w/ custom header: Header does not look custom
    - subheader: should we put something more typical in the subheader...it doesn't exactly look _designed_
    - with drawer layout: the rail is not hiding the footer content by default...this may be a bug in the implementation
- InfoListItem
    - with icon: leaf is not a status icon...swap for alarm?
    - with array subtitle: text and icons don't really go together at all
    - with background color: default to white?
    - with left component: use a timestamp instead (most typical thing that would go there)
    - with full config: use a different default status color
    - in a full list: leaf is not an 'online' icon
- ListItemTag
    - custom colors: maybe start off with red/white
    - full config: start with default colors
    - in ILI: use red[500] for bugs

emclaug2:
> `startIcon`

I'm not sure what this `startIcon` prop is.  Is it something that goes on the AddIcon directly?

joebochill:
> > `startIcon`
> 
> I'm not sure what this `startIcon` prop is. Is it something that goes on the AddIcon directly?

https://material-ui.com/components/buttons/#buttons-with-icons-and-label

emclaug2:
`with custom header` story should look like this: 

image

emclaug2:
>   * with drawer layout: the rail is not hiding the footer content by default...this may be a bug in the implementation

Component implementation looks fine to me, the story was updated to hide the footer if the variant is a rail.

joebochill:
> `with custom header` story should look like this:

Yeah, that's what I'm seeing. But the intent of 'custom header' is to use it when you can't achieve what you need with the basic props. This is just a title and subtitle, which is easily achieved with the respective props. It might be more useful to show something more custom, like an extra icon or a badge or version tag or something that actually requires using the custom header implementation.

emclaug2:
> > `with custom header` story should look like this:
> 
> Yeah, that's what I'm seeing. But the intent of 'custom header' is to use it when you can't achieve what you need with the basic props. This is just a title and subtitle, which is easily achieved with the respective props. It might be more useful to show something more custom, like an extra icon or a badge or version tag or something that actually requires using the custom header implementation.

This example isn't using the basic props of the Drawer;  the header is all custom content.  The title is smaller than the subtitle.   We can adjust this to be more custom like throwing in an icon somewhere but I think it'll be hard to read because of the current background image.

joebochill:
> This example isn't using the basic props of the Drawer; the header is all custom content. The title is smaller than the subtitle. We can adjust this to be more custom like throwing in an icon somewhere but I think it'll be hard to read because of the current background image.

I know it's not _actually_ using the props...but if you toggle between the full config story (which is) and the custom header, the difference is barely noticeable. What if you just pop a ListItemTag in there: 

```
Custom Header               | v1.0.0 |
Content 
```

emclaug2:
Ok all the changes should be pushed up now; I'll start working on the angular changes.

emclaug2:
Hosted here: https://pxblue-react-library.web.app/?path=/story/intro-overview--px-blue-react-components

emclaug2:
> image
> note the font weight difference in the body list

If this is from the showcase I'll make a separate issue for this item.  Showcase will need an overhaul anyways to add the kitchen sink.

huayunh:
The picture comes from DSM

joebochill:
**Drawer with nested list items**: divider off by default
**Hero within a hero banner**: swap out leaf icon for a battery
**ILI with background**: default backgroundColor to white, consider changing leaf icon
**ILI with rightComponent**: right component can use the channelValue from before (right side is typically for channel values, left side is typically for timestamps)
**ILI within full list**: replace the leaf icon for status (left icon) and temperature (right icon) rows

Otherwise looks good. I think Huayun is talking about body of the scorecards...the lists there are using the ILI component which uses a semibold title by default, but she's saying they should be regular. That will require an override using a class (`title`)

huayunh:
image
In Drawer's "within a drawer layout" story, would you change the custom drawer header title to also include a right padding of 16px for the sake of i18n?

huayunh:
Adjust the custom title block's padding so that it is always centered, as the Drawer Header changes to 56px on mobile. 

image

huayunh:
Propose to change the second drawer group to those nav items: 
image

(those two represent a more common use case)

huayunh:
I propose getting rid of the nav group title in 
https://pxblue-react-library.web.app/?path=/story/components-drawer--with-nested-list-items
since there is only one nav group

huayunh:
In https://pxblue-react-library.web.app/?path=/story/components-drawer--with-full-config, the two nav group's typographies are not matching with each other properly. Should both use overline — at least for the group title part. Happy either way with the version number.

huayunh:
> **Drawer with nested list items:** divider off by default

I am cool with leaving the dividers on in this case, because it is easier to tell the nav hierarchy.

huayunh:
https://pxblue-react-library.web.app/?path=/story/components-dropdown-toolbar--with-custom-menu
Those are not icons with directionality and thus should not get flipped in RTL

image

huayunh:
image

Change to something more sense-making...
"No devices"
"Check your network connection or add a new device"

huayunh:
https://pxblue-react-library.web.app/?path=/story/components-hero--with-icon-colors 

Icon color default to white50

huayunh:
https://pxblue-react-library.web.app/?path=/story/components-info-list-item--with-array-for-subtitles
Default separator to interpunct `·`

huayunh:
https://pxblue-react-library.web.app/?path=/story/components-info-list-item--with-left-component

Timestamp shall have different font-weight. Here is a screenshot from DSM:
image

huayunh:
Aside from what Joe's saying and that user menu font color problem, everything else appears good

emclaug2:
> > **Drawer with nested list items:** divider off by default
> 
> I am cool with leaving the dividers on in this case, because it is easier to tell the nav hierarchy.

What about nested dividers?

emclaug2:
Screen Shot 2021-01-20 at 1 17 08 PM

How's this?

huayunh:
> What about nested dividers?

I prefer off

emclaug2:
All audit items should be addressed at this point, now working on angular updates.  

See updated storybook here: https://pxblue-react-library.firebaseapp.com/?path=/story/*

huayunh:
https://pxblue-react-library.firebaseapp.com/?path=/story/components-drawer--with-custom-header

image

huayunh:
https://pxblue-react-library.firebaseapp.com/?path=/story/components-drawer--with-nested-list-items 
Icons for "user guide" and "community" may be changed to
* `chrome_reader_mode` or `menu_book`
* `groups`

huayunh:
image
this guy still here

emclaug2:
> image
> this guy still here

I have no idea how I missed this 3x.  Fixing this first.

huayunh:
image
logout icon shall be flipped under RTL

huayunh:
> I have no idea how I missed this 3x. Fixing this first.

no worries, I didn't catch all the design flaws in one go either :P

emclaug2:
> image
> logout icon shall be flipped under RTL

This is more complicated than expected since I don't think our RTL util function is working as expected.  I'm making a separate issue for this one since I can't fix it in <5 minutes.

emclaug2:
Ok drawer custom header, nested nav icons, and user menu crazy colors should be fixed now. 

https://pxblue-react-library.firebaseapp.com/?path=/story/*

huayunh (approved):
https://pxblue-react-library.firebaseapp.com/?path=/story/components-score-card--with-score-badge
image
Flip list icon

https://pxblue-react-library.firebaseapp.com/?path=/story/components-hero--with-full-config
image
flip trending icon

joebochill:
Looks good...last thing: we're still using this leaf icon for temperature in the full list story:
image
 And you may want to take a new screenshot for the readme if you like:
image

huayunh:
Ha, I was just about the suggest that! Evan can you make a story to update all the screenshots in storybook so that they are
(1) Hi-Res
(2) Reflect the proper design

For each component, the readme shall include at least two images, one for the component itself, the other for how it is used in its context?

emclaug2:
Latest RTL changes are pushed out now, new site can be found here: https://pxblue-react-library.web.app

I made a story to update the component docs with the latest images.   Thank you both for the detailed audit!

joebochill (approved):

Issue #216, opened 2021-01-07
Duplicated classes on Drawer items


joebochill:
#### Describe the bug
Some classes are being added multiple times to the Drawer Item titles

#### What is the expected behavior?
Classes should only be applied once without duplication

#### What are the steps to reproduce?
1. Launch the showcase demo
2. Inspect the title of an item in the Nav Drawer (the span)
3. Observe all of the classes are there twice (or more)

#### Screenshots or links to minimum reproduction example
![image](https://user-images.githubusercontent.com/29152776/103919310-e41c8180-50dd-11eb-849c-15950541d72c.png)

		
#### Anything else to add?
I messed with this for a minute and couldn't figure it out. The props coming in to the NavItem seem to only have them listed once, but idk. Weird.

emclaug2:
@joebochill might fix this as part of the refactor story.

joebochill:
I came cross an issue on MUI that seemed to indicate that this might be a problem on their end (that they say is fixed in v5) related to nested style rules in JSS that depend on props. I can't seem to find that issue now, but if I do I will link it.

joebochill:
Let's see if this is fixed after we move to v5.

emclaug2:
We're now on MUI5; if this issue isn't needed it can be closed.

joebochill:
Need to assign somebody to check this.

daileytj:
These classes are still being duplicated

joebochill:
This only seems to be an issue with the InfoListItem / DrawerNavItem. I thought it might have something to do with the ListItemButton conditional wrapper for clickable items, but removing that still shows some duplicated classes (e.g., BLUIDrawerNavItem-title).

joebochill:
The issue is in our use of the `combine` function. It combines the `defaultClasses[xxx]` with `classes[xxx]`, but `defaultClasses` already accounts for the classes prop.

<img width="849" alt="image" src="https://github.com/etn-ccis/blui-react-component-library/assets/29152776/6ab5e03b-f672-44a6-a79b-8f8fee8b43b8">
<img width="790" alt="image" src="https://github.com/etn-ccis/blui-react-component-library/assets/29152776/8c6f0943-aaf8-44f9-8c51-9ba43e02ddaf">

This is a problem in every component...we're just seeing it in the DrawerNavItem because it's the only place where we have actually used the `classes` prop for one of our components.

Confirmed this by using `classes` on the ChannelValue demo:
[...]

PR #230, opened 2021-01-22
Drawer Refactor

### LINKED ISSUE #139: Streamline Inheritance for Drawer Props
Try to make the inheritance of Drawer -> DrawerBody -> DrawerNavGroup -> DrawerNavItem props less verbose by adding a function to handle inheritance.  React Native has an example of this. 

Each component independently handles inheritance.

No

### LINKED ISSUE #221: Class inheritance in the drawer is goofy
#### Describe the bug
The DrawerNavItem inherits `classes` from the DrawerNavGroup, but the classes used in the NavItem and the NavGroup are different. This can cause some unexpected behaviors or warnings in the console when trying to do class overrides.

I would like to see this refactored such that the DrawerNavGroup takes a `classes` prop of type DrawerNavGroupClasses, which only apply to the DrawerNavGroup itself. And then add a new prop to DrawerNavGroup (maybe another prop that can be passed in at the Drawer, DrawerBody, or DrawerNavGroup level) for `navItemClasses` that get passed down to all of the NavItem children.

#### Screenshots
NavGroup Classes and type:
image

NavItem classes:
image

		
#### Anything else to add?
I think this is only an issue in the Drawer because of how we have implemented the prop/style inheritance through the component hierarchy, but it may be worth looking to make sure that there isn't a similar issue with any other components.

The types of DrawerRailItem and DrawerNavItem are also slightly different. The DrawerNavGroup `items` prop should be able to support either...this will probably need some typescript magic.

joebochill:
For PXBLUE-1808.
Fixes #221 
Fixes #139 

Ladies and gentlemen, here it is...the long awaited and highly-anticipated refactor or the drawer component! 

I did some testing on this to make sure things still work as expected, but would like some eyes to give it a deep and thorough run through to make sure I didn't miss anything.

- Refactor the drawer sub components to be able to render as standalone items
    - Each item appearance is determined completely deterministically from its props
- Shared/Inherited prop types were cleaned up and separated to accommodate rail items and nav items
- Added common items to the drawer context (activeItem, etc.) to avoid having to pass them through the hierarchy as props.
- Cleaned up all of the style/class overrides and make sure we are applying them and exposing them correctly
- Moved the nesting logic into the DrawerNavItem instead of the DrawerNavGroup

If you so choose, you can now build a drawer by passing in children instead of using the items array (or you can even mix and match if you want), e.g.:

```
<Drawer>
    <DrawerHeader />
    <DrawerBody>
        <DrawerNavGroup>
            <DrawerNavItem>
                <DrawerNavItem>
                    <DrawerNavItem/>
                </DrawerNavItem>
            <DrawerNavItem>
            <DrawerNavItem/>
        </DrawerNavGroup>
    </DrawerBody>
    <DrawerFooter />
</Drawer>
```

The drawer should continue to look and function exactly as it did before (maybe slightly better) — visually you shouldn't see any difference.


huayunh:
Love the new structure! I will take a look at this on Monday or Tuesday

emclaug2:
I'm taking a look right now; one thing I noticed was that the `full config` drawer story has a duplicate divider in the DrawerFooter.  Could you remove it so we only use the prop?

joebochill:
> I'm taking a look right now; one thing I noticed was that the `full config` drawer story has a duplicate divider in the DrawerFooter. Could you remove it so we only use the `divider` prop?

Consider it removed. I'll just sit on pushing it to see if there are other change requests I can bundle with it.

emclaug2:
DrawerBody line 40 you can remove the eslint comment.  

`// eslint-disable-line @typescript-eslint/no-unused-vars`

emclaug2:
Nice mergeStyleProp function, this is looking a lot cleaner.

joebochill:
> Nice mergeStyleProp function, this is looking a lot cleaner.

Pretty sure I stole this from what you did for RN 😆 

I wanted to make it even more powerful (less repetitive) by letting you pass in an array of props to merge so we could do the merge in a single line, but it got kinda hairy so I left it the way it is for now.

emclaug2:
> DrawerBody line 40 you can remove the eslint comment.
> 
> `// eslint-disable-line @typescript-eslint/no-unused-vars`

Same applies to line 117 of DrawerNavGroup
` // eslint-disable-line @typescript-eslint/no-unused-vars`

emclaug2:
DrawerNavItem line 501 and 460 @ts-ignore can be removed as well; seems line 481 still gives issues.

emclaug2:
I really like this refactor; I think this makes more sense than before, it's more flexible for users.  Only stuff I noticed was some lint comments and @ts-ignore comments that were probably outdated.

emclaug2 (approved):

huayunh:
Is it possible to make the `title` of DrawerRailItem optional and have it default to an empty string?

huayunh:
Nested divider is on when the knob turns it off
http://localhost:6006/?path=/story/components-drawer--with-nested-list-items
image

huayunh:
Nav Rail, check the `condensed` knob

image
http://localhost:6006/?path=/story/components-drawer--with-different-variants

huayunh:
Can we rename Accordions so that the console won't scream out red texts?
image
http://localhost:6006/?path=/story/components-drawer--with-subheader

huayunh:
It looks like when `nestedBackgroundColor` is set, the `activeItemBackgroundColor` is gone. I expect the activeItemBackgroundColor to be an overlay on top of the `nestedBackgroundColor`
image

http://localhost:6006/?path=/story/components-drawer--with-nested-list-items

huayunh:
Same for the regular backgroundColor

huayunh:
Full config story, can we add a "(persistent only)" to the knob name "openOnHover"?

huayunh:
I added `classes={{ root: 'root-test' }}` to line 319 of `react-component-library/demos/storybook/stories/drawer/with-full-config.tsx`, and two classes show up
image

huayunh:
RTL indentation on depth 2 items
image

http://localhost:6006/?path=/story/components-drawer--with-nested-list-items

huayunh:
Design and code otherwise look fine... my TODO: read README

huayunh:
Really like how you handle the prop inheritance

joebochill:
> Is it possible to make the `title` of DrawerRailItem optional and have it default to an empty string?

We talked about this on the original drawer rail story. I'm inclined to leave it for now (we don't want people to use a regular rail without labels do we?). We're using it for the tooltips on the condensed style as well.

huayunh:
> we don't want people to use a regular rail without labels do we?

I personally think this falls into that "technically possible but we don't encourage you to do so from the design perspective"

joebochill:
> It looks like when `nestedBackgroundColor` is set, the `activeItemBackgroundColor` is gone. I expect the activeItemBackgroundColor to be an overlay on top of the `nestedBackgroundColor`

It's there...it's blue[500] at 5% opacity on black so it's nearly impossible to see.

joebochill:
> I added `classes={{ root: 'root-test' }}` to line 319 of `react-component-library/demos/storybook/stories/drawer/with-full-config.tsx`, and two classes show up

There is a separate bug logged for the duplicate classes: #216

joebochill:
> RTL indentation on depth 2 items

This is currently broken in dev (i.e. not related to the refactor). We may want a separate issue for this if I can't find a quick fix for it.

huayunh:
Write unit tests for the declarative styled components

joebochill:
@huayunh I think I covered all of your comments. If not let me know.

CHANGELOG.md
```diff
@@ -1,11 +1,15 @@
 # Changelog
 
-## version tbd (not published)
+## 5.0.0 (not published yet)
```
huayunh, line 3: We don't have anything broken, just a big refactor on drawer, right? Asking this to ensure that we don't need a `_breaking changes_` section

docs/Drawer.md
```diff
@@ -263,176 +280,184 @@ Any other props supplied will be provided to the root element (`div`).
...
+| rightComponent                  | An icon/component to display to the right                                                                                                                   | `ReactNode`             | no       |         |
+| statusColor                     | Status stripe and icon color                                                                                                                                | `string`                | no       |         |
+| subtitle                        | The text to show on the second line                                                                                                                         | `string`                | no       |         |
+| title                           | The text to show on the first line                                                                                                                          | `string`                | yes      |         |
```
huayunh, line 308: `title` is now no longer required, `icon` is

components/src/core/Drawer/DrawerRailItem.tsx
```diff
@@ -24,48 +30,36 @@ export type DrawerRailItemClasses = {
...
-        | 'activeItemFontColor'
-        | 'activeItemIconColor'
-        | 'divider'
-        | 'itemFontColor'
-        | 'itemIconColor'
-        | 'onItemSelect'
-        | 'ripple'
-    > & {
-        activeItem?: string;
+export type ExtendedNavItem = NavItem & { ButtonBaseProps?: Partial<MuiButtonBaseProps> };
```
huayunh, line 33: Right now if I try to do 
```tsx
<DrawerNavGroup items={[{ icon: <></>, itemID: 'asdf' }]} />
```
it will fail, but if I do 
```tsx
<DrawerNavGroup>
    <DrawerRailItem itemID={''} icon={<></>} />
</DrawerNavGroup>
```
It will work.

Maybe we can do some console warning tricks, or insert a blurp about this in the readme.

components/src/core/Drawer/DrawerRailItem.tsx
```diff
@@ -276,3 +279,27 @@ const DrawerRailItemRender: React.ForwardRefRenderFunction<unknown, DrawerRailIt
...
+        titleActive: PropTypes.string,
+        ripple: PropTypes.string,
+    }),
+    hidden: PropTypes.bool,
+    icon: PropTypes.element,
+    itemID: PropTypes.string.isRequired,
+    onClick: PropTypes.func,
+    statusColor: PropTypes.string,
+    title: PropTypes.string.isRequired,
+    // @ts-ignore
```
huayunh, line 303: icon required, title not required

huayunh:
Since it is only for the drawer and no other components, I must say I strongly dislike the idea of moving the selected logic into ILI. It just gets creepy when you can do a "rounded" highlight shape in a list item.

I would still prefer a 
```tsx
<div style={{backgroundColor: props.backgroundColor}}>
    <activeHighlight style={{backgroundColor: props.activeItemBackgroundColor}} />
    <InfoListItem />
</div>
```
structure

CHANGELOG.md
joebochill: There are breaking changes (class name changes and moving the activeItem prop to the drawer instead of the navgroup are the most breaking), hence the major version bump. The Keepachangelog format we use does not have a separate section for breaking changes

docs/Drawer.md
joebochill: That's backwards. For the NavItem, the title is required and icon is optional. For RailItem, they are both optional.

components/src/core/Drawer/DrawerRailItem.tsx
joebochill: I updated the type for items to take an array of navItems or railItems, so you should be able to do either of these forms now.

CHANGELOG.md
joebochill: I could potentially do a little work to allow you to continue to specify activeItem on the nav group to maintain backwards compatibility, but I would prefer not to (originally we had posed that doing it that way would allow people to have multiple items active at a time in different groups, but I don't think that's really a use case we need to support).

components/src/core/Drawer/DrawerRailItem.tsx
huayunh: It gets cast to the InfoListItem level and I am still getting yelled at. I was expecting:

### Pass
```tsx
        <Drawer variant={'rail'}>
            <DrawerBody>
                <DrawerNavGroup items={[{ icon: <></>, itemID: 'b' }]} />
                <DrawerNavGroup>
                    <DrawerRailItem icon={<></>} itemID={'a'} />
                </DrawerNavGroup>
            </DrawerBody>
        </Drawer>
```

### Fail (or error thrown)
```tsx
        <Drawer variant={/* some other variant */}>
            <DrawerBody>
                <DrawerNavGroup items={[{ icon: <></>, itemID: 'b' }]} />
                <DrawerNavGroup>
                    <DrawerRailItem icon={<></>} itemID={'a'} />
                </DrawerNavGroup>
            </DrawerBody>
        </Drawer>
```

### Pass
```tsx
        <Drawer variant={/* some other variant */}>
            <DrawerBody>
                <DrawerNavGroup items={[{ title: 'b', itemID: 'b' }]} />
                <DrawerNavGroup>
                    <DrawerRailItem title={'a'} itemID={'a'} />
                </DrawerNavGroup>
            </DrawerBody>
        </Drawer>
```

### Fail (or error thrown)
```tsx
        <Drawer variant={'rail'}>
            <DrawerBody>
                <DrawerNavGroup items={[{ title: 'b', itemID: 'b' }]} />
                <DrawerNavGroup>
                    <DrawerRailItem title={'a'} itemID={'a'} />
                </DrawerNavGroup>
            </DrawerBody>
        </Drawer>
```

joebochill: Fixing number 1 (it's just a small issue with the PropTypes definition). Numbers 2 and 3 are not valid use cases...you shouldn't be using a RailItem inside of a drawer that is not the rail variant...i don't see us supporting that.

4 is failing because you just told me to make the icon prop required for a DrawerRailItem and you haven't provided one.

joebochill: I can go back to making the icon optional if that's what we want to do...the rail item can functionally handle it either way.

huayunh: >  RailItem inside of a drawer that is not the rail variant

My bad for copy pasting. Corrected

joebochill: Okay, so 2 is failing because you don't have a `title` prop which is required for NavItems. And 4 is failing because you don't have an `icon`, which is required for a RailItem.

huayunh: Yup, that's the expected behavior. I see that you made more changes... will review during the OH

huayunh (approved):
🔥

PR #243, opened 2021-02-19
Update components to work with version 6 of the theme

joebochill:
For PXBLUE-XXXX

- Update Drawer Header, Scorecard Header, and UserMenu to reflect the updates to version 6 of the themes

To test, you'll need to pull down the latest themes, build them, and manually copy them into the storybook or showcase for the time being.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 5.0.1 (Not published yet)
+
+### Changed
+
+-   Some default colors in `<Drawer>` and `<ScoreCard>` headers to work with @pxblue/react-themes version 6+.
+
 ## 5.0.0
 
 ### Changed
--- a/README.md
+++ b/README.md
@@ -33,6 +33,10 @@ The library can be built by running the following command. The resulting output 
 yarn build
 ```
 
+## Using with @pxblue/react-themes
+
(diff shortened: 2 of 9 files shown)


emclaug2:
I'm seeing no distinction between nested nav items & nav items in the drawer: 

image

joebochill:
> I'm seeing no distinction between nested nav items & nav items in the drawer:

Did you build manually build the latest theme changes and copy them into storybook node modules? 

Should look like this:
image

emclaug2:
@joebochill  You were right, I didn't manually build the theme; Drawer and Scorecard look fine to me, but ListItemTag story is giving some trouble now: 

image

joebochill:
> @joebochill You were right, I didn't manually build the theme; Drawer and Scorecard look fine to me, but ListItemTag story is giving some trouble now:

There are a number of places I saw where storybook breaks down going from light to dark theme because of hard coded colors that are only suitable for light theme. 

This one was hard coded to a white background, which I thought I swapped to using the paper background color from the theme but it must not have taken. 

It’s probably worth logging an issue to address storybook issues when switching themes. I’d rather not try to tackle them all here if I can help it since it’s not directly tied to the theme changes.

emclaug2 (approved):

joebochill:
I logged #244 to capture issues when toggling the theme

PR #246, opened 2021-03-01
Update typography variants for DrawerHeader and Hero components

daileytj:
- Update typography variants for DrawerHeader and Hero components

### DIFF
--- a/components/src/core/Drawer/DrawerHeader.tsx
+++ b/components/src/core/Drawer/DrawerHeader.tsx
@@ -100,7 +100,6 @@ const useStyles = makeStyles((theme: Theme) =>
             marginLeft: theme.spacing(1),
         },
         subtitle: {
-            fontWeight: 300,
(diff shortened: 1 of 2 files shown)


components/src/core/Drawer/DrawerHeader.tsx
```diff
@@ -100,7 +100,6 @@ const useStyles = makeStyles((theme: Theme) =>
             marginLeft: theme.spacing(1),
         },
         subtitle: {
-            fontWeight: 300,
```
joebochill, line 103: Font-weight 300 is light, I believe. Are we using regular (400) in the DSM?

daileytj: Yeah, it's 400 in the DSM. I couldn't find a formal design for the DrawerHeader, but the appbar pattern has 400 font-weight for the subtitle and Huyaun confirmed that it is using Body2 which is 14px 400 font-weight.

huayunh:
I have a feeling that when we redesign & completely lock down the design of the appbars, some of the efforts here will need to be redone...

PR #252, opened 2021-03-19
Feature/rem system sizing

emclaug2:
-  Use Rem styles so that the font-size & vertical spacing scales with system font size. 
- Dropdown Toolbar header-spacing still imperfect.


emclaug2:
Hosted here: https://pxblue-react-library.web.app/

huayunh:
image
On extra small, those text did not line up... I personally am not particularly concerned about extra small size, but just wanted to point this out since Angular side had this correct.

huayunh:
Extra large, in https://pxblue-react-library.web.app/?path=/story/components-drawer--with-different-variants, the header overflowed
image

huayunh:
And may we bump up the theme to the latest beta `^6.0.0-beta.1` to pick up the changes to the overline typographies... just to see how those integrate

joebochill (commented):
Changes look reasonable. Do we need to update anything in the showcase similar to what you've done for the storybook inline font sizes?

joebochill:
1. Collapsed drawer looks broken for me with extra-large font:
image

2. Rail Header icon alignment is off:
image

3. Should we be converting the iconSize for Hero into rems behind the scenes? I'm not fully decided on how i feel about it yet.

4. Top and bottom spacing on the ScoreCard doesn't look even...maybe because of the inexact conversion of 100px to rems:
image

5. User Menu avatars:
image

6. This looks off...should the toolbar be taller?
image

emclaug2:
> Should we be converting the iconSize for Hero into rems behind the scenes? I'm not fully decided on how i feel about it yet. 

Since our inputs accept pixels I don't think we should do any conversion here.  Drawer width in angular-component-lib is maintained in pixels.

emclaug2:
- The scorecard header spacing has been translated to 100px, not 96px. 
- Drawer closed items are center aligned now. 
- Toolbar in UserMenu context story has a rem height now. 

Some things I'm having some trouble with:
- UserMenu avatar not center-aligned.  Even though the Menu Icon is center aligned, the avatar doesn't want to get out of the corner.  
- Drawer Header icon margin spacing.
- Dropdown Toolbar icon spacing.

joebochill:
> Thanks for the comments @huayunh & @joebochill
> 
> Here's what been re-deployed:
> 
> * The scorecard header spacing has been translated to 100px, not 96px.
> * Drawer closed items are center aligned now.
> * Toolbar in UserMenu context story has a rem height now.
> 
> Some things I'm having some trouble with:
> 
> * UserMenu avatar not center-aligned.  Even though the Menu Icon is center aligned, the avatar doesn't want to get out of the corner.
> * Drawer Header icon margin spacing.
> * Dropdown Toolbar icon spacing.

@emclaug2 it looks like you didn't push any of these updates to your branch, which makes it challenging for me to finish this up for you.

joebochill:
I managed to recreate your fixes (i think) by digging into your deployed storybook and using the inspector. Working through the couple leftover items now.

joebochill:
Okay, I think I have addressed the items that Evan didn't quite finish. I'm cool with this PR now, but if someone else wants to take a look for sanity checking, that would be great.

huayunh:
I expect the drawer header to be influenced by rem as well.
image

huayunh:
image
not aligned

huayunh:
Screen Shot 2021-03-24 at 3 27 37 PM

huayunh:
Screen Shot 2021-03-24 at 3 29 36 PM

huayunh:
Extra small, user menu 
image

huayunh:
expand / collapse icon seem not working in drawer. Not sure if it is from this PR.

huayunh:
InfoListItem, with left component, RTL
image

joebochill:
Very strange...because it looks fine in the showcase:
image

I'll have a look and see what's going on with storybook tomorrow.

joebochill:
Not sure about the expand icons not working...they work in the showcase and storybook for me...give it another shot?

huayunh (approved):
Nice! 

The only thing that catches my eye is the toolbar in http://localhost:6006/?path=/story/components-user-menu--within-toolbar, but I suppose this min-height comes from the theme definition and is not controlled by us?

Actually one more thing: I think the round shape's round corner shall be set to `50%`, not a fixed pixel value
image

And a theme thing (it never ends!)
image
I believe the active subtitle shall also turn blue200 in the dark theme

joebochill:
> Nice!
> 
> The only thing that catches my eye is the toolbar in http://localhost:6006/?path=/story/components-user-menu--within-toolbar, but I suppose this min-height comes from the theme definition and is not controlled by us?

Yes, the min-height comes from the default Material rules. We could override it in our themes, but I think that's probably overkill given the low likelihood of somebody wanting to use a smaller than normal font. Our Dropdown toolbar handles the size appropriately in REM. For this example, I just set the minimum inline in the story.

> Actually one more thing: I think the round shape's round corner shall be set to `50%`, not a fixed pixel value

Fun bit of trivia, if you've never run into it before — `border-radius` as a percentage is actually based on the _width_ of an element, not the height. So in instances like this, with a rectangle, we actually must specify a numerical value, or you get something like this:
image

> And a theme thing (it never ends!)
> I believe the active subtitle shall also turn blue200 in the dark theme

You are correct. This happened when we changed the subtitle to use textSecondary...didn't account for the special case of active.


Fixes incoming.

joebochill:
@huayunh I noticed some misalignment in the Drawer Rail header when using a regular icon vs. a clickable button icon, so I fixed that. Your findings also prompted me to do a full audit of the component code. I found a few more small items that were missed (e.g. negative margins in ListItem text subtitles, LIT border radius, etc.) so I cleaned those up. You may want to take one final quick pass through to make sure nothing got broken.

huayunh:
The only thing I noticed is that

image

the spacing between timestamp and titles are different RTL and LTR, because 

image

seems to have a constant left margin

huayunh:
Can you use a hook to change the ILI background color for 
http://localhost:6006/?path=/story/components-list-item-tag--in-info-list-item ?

huayunh:
In the cover page (http://localhost:6006/?path=/story/intro-overview--px-blue-react-components), can you render the GitHub logo as a `startIcon` prop?

huayunh:
In http://localhost:6006/?path=/story/components-drawer--with-subheader, maybe you'd want to do the same for that search icon... and perhaps change the subtitle padding to `1rem 16px`

huayunh:
For http://localhost:6006/?path=/story/components-info-list-item--within-full-list, if possible, would be better if avatar bg and  status stripe is red500, while the texts stay 200 in dark thee

huayunh:
No other problem spotted

joebochill:
Since these are storybook issues and not directly related to the REM sizing changes, I'm going to log a separate issue for that and handle them in a separate PR.

PR #254, opened 2021-03-20
Update empty state description color

### LINKED ISSUE #253: Update Empty State description text color to black
Description text of the empty state color should be black (ask Huayun which value)

Currently uses blue[500].

No

#### Additional Context

image

joebochill:
Fixes #253 .

- Switch the description color to text primary (light) or text secondary (dark)

#### Screenshots:
image
image

### DIFF
--- a/components/src/core/EmptyState/EmptyState.tsx
+++ b/components/src/core/EmptyState/EmptyState.tsx
@@ -36,6 +36,9 @@ const useStyles = makeStyles((theme) => ({
         display: 'flex',
         fontSize: 96,
     },
+    description: {
+        color: theme.palette.type === 'dark' ? theme.palette.text.secondary : theme.palette.text.primary,
+    },
     actions: {
         marginTop: theme.spacing(1),
     },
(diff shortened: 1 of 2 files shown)


jeffvg (approved):

components/src/core/EmptyState/EmptyState.tsx
```diff
@@ -54,7 +58,11 @@ const EmptyStateRender: React.ForwardRefRenderFunction<unknown, EmptyStateProps>
                 {title}
             </Typography>
```
huayunh, line 59: What's the color of the title here? Eyeballing from the screenshot it looks like textSecondary to me. Should be textPrimary in both light and dark.

joebochill: The default typography is inherit, so it inherits from root, which is textPrimary (see line 26). The demo is using an outdated version of the theme. With the latest it looks like:

image
image

huayunh: Then visually that screenshot looks awesome :)

PR #272, opened 2021-04-05
Publish 5.1.1

huayunh:
- Fixed a minor alignment issue in `<DrawerHeader />`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 5.1.1
+
+### Fixed
+
+-   Fixed misaligned menu icon in `<DrawerHeader>`.
+
 ## 5.1.0
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 -   Some default colors in `<InfoListItem>`, `<Drawer>` and `<ScoreCard>` headers to work with @pxblue/react-themes version 6+.
 -   Some default styles on `<ScoreCard>`, `<UserMenu>`, and `<EmptyState>` text elements.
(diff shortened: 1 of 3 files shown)


joebochill:
So this is caused because in Safari, the 100% height of the `navigation` is not respected because something about a flex child in a flex parent without explicit height percentages not working. 

Anyway, do we want to fix this by forcing all of the DrawerHeader content to be center aligned (current fix), or my adjusting the navigation element to be `alignSelf: flex` instead of `height: 100%`? I'm wondering specifically about the case when a user provides their own custom content. Should it be centered by default, or should we give them a blank canvas to work from, e.g.:

image

I think centering is probably fine and the more likely direction people will go in for custom contents anyway, but just wanted to ask.

joebochill:
On an unrelated related note, this is a good opportunity to bring up the idea of version planning again too. We don't want to cram everything into one release and only do one per quarter, but we also probably don't want to do a new patch release for every bug fix either. We should probably try to knock out a few bugs in each patch, generally.

In this case, I think it's fine since there aren't any other bugs in the library right now. It might be worth waiting to see if any more are discovered in 5.1.0 since it's only been out for 4 days. 🤷 either way works for me. Just food for thought.

emclaug2:
> Anyway, do we want to fix this by forcing all of the DrawerHeader content to be center aligned (current fix), or my adjusting the navigation element to be `alignSelf: stretch` instead of `height: 100%`? I'm wondering specifically about the case when a user provides their own custom content. Should it be centered by default, or should we give them a blank canvas to work from, e.g.:


I think centering the content is a reasonable default; if they require a design or some wacky content that needs to be off-centered they can just override the styles.

joebochill (approved):

daileytj (approved):

PR #277, opened 2021-04-23
Add publish script

emclaug2:
- Add publish script

### DIFF
--- a/PUBLISHING.md
+++ b/PUBLISHING.md
@@ -13,7 +13,6 @@ If you need to publish a package manually, you can run the following commands fr
 
 ```sh
 yarn build 
-cd dist
 yarn publish:package
 ```
 
@@ -23,7 +22,6 @@ For this command to work, you must have an NPM token configured in your environm
 
 ```sh
 yarn build 
-cd dist
 npm adduser && yarn publish:package
 ```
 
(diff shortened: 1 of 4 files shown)


joebochill (commented):
Code looks good...just see my comment about reorganizing the publishing.md file.

PUBLISHING.md
```diff
@@ -1,13 +1,30 @@
...
+  In both cases, the code will only be published if the version number differs from the current version published under the respective dist tag.
 
+## Manually Publishing
+
+If you need to publish a package manually, you can run the following commands from the root folder:
+
+```sh
+yarn build 
+cd dist
+yarn publish:package
```
joebochill, line 17: Same comment here about moving the cd into the publish command

emclaug2: This one should work as-is because the scripts section isn't removed as part of the build.  I can make all 3 component libs have the same top-level publish command for consistency though.

joebochill: That way everything will look the same in the CI config file and the publishing doc

emclaug2: Ok RN and React libraries match now.

joebochill (approved):

PR #281, opened 2021-05-07
Refactor drawer story state and update drawer footer image for dark theme

### LINKED ISSUE #249: Update drawer footer example for dark theme
#### Describe the bug
The drawer foot needs updated to match what we have in DSM. Currently the Eaton Logo font is not readable and the Eaton logo color should be swapped

image

daileytj:
Fixes #249.

- Refactor state for drawer stories
- Update footer images for use with theme toggle

*Notes:
- The footer image swaps for the with-footer and full-config stories but there is a knob to change the background color so it's "hardcoded" when the theme is toggled. Should we still swap out the image even though we can't change the "hardcoded" color value? Or is there a way to make a knob default dynamic so we can hardcode the color with respect to the theme?
- I'm getting a hooks issue when toggling to the rail variant on the within-a-drawer-layout story :( ... still need to figure that out.


demos/storybook/package.json
```diff
@@ -12,7 +12,7 @@
         "@pxblue/react-themes": "^6.0.0",
         "@pxblue/storybook-rtl-addon": "^1.0.0",
         "@pxblue/storybook-themes": "^1.1.0",
-        "@sambego/storybook-state": "^1.3.6",
+        "@sambego/storybook-state": "^2.0.0",
```
joebochill, line 15: If we're refactoring to local state, I don't think we need this package at all anymore

demos/storybook/stories/drawer/with-basic-config.tsx
```diff
@@ -1,54 +1,57 @@
...
-    },
-    {
-        title: 'Notifications',
-        itemID: '4',
-        icon: <NotificationsActive />,
-    },
-];
+const store: Store<DrawerState> = new Store<DrawerState>({
+    selected: '',
+});
```
joebochill, line 13: This doesn't actually do anything...you never read it and never change it (you're just using the local react state variable below).

demos/storybook/stories/drawer/with-basic-config.tsx
```diff
@@ -1,54 +1,57 @@
...
+            icon: <Accessibility />,
+            onClick: () => setSelected('3'),
+        },
+        {
+            title: 'Notifications',
+            itemID: '4',
+            icon: <NotificationsActive />,
+            onClick: () => setSelected('4'),
+        },
+    ];
```
joebochill, line 43: You can make these menu items a shared function so you don't have to copy/paste/redefine them in every story.

demos/storybook/stories/drawer/with-footer.tsx
```diff
@@ -1,31 +1,82 @@
...
+                    hideContentOnCollapse={boolean('hideContentOnCollapse', true)}
+                    divider={boolean('divider', true)}
+                >
+                    <div style={{ display: 'flex', justifyContent: 'center', flexDirection: 'row' }}>
+                        <img
+                            src={useDarkMode() ? EatonFooterLogoDark : EatonFooterLogoLight}
+                            style={{
+                                margin: 16,
+                                marginRight: 30,
+                                marginLeft: 8,
```
joebochill, line 63: These margins seem kinda goofy — what are we trying to do here?

joebochill:
> * The footer image swaps for the with-footer and full-config stories but there is a knob to change the background color so it's "hardcoded" when the theme is toggled. Should we still swap out the image even though we can't change the "hardcoded" color value? Or is there a way to make a knob default dynamic so we can hardcode the color with respect to the theme?

I haven't tried this before, but you could try using a `useEffect` hook to respond to a change to the `useDarkMode` value and then redeclare the knob with a new default value:

```
const darkMode = useDarkMode();
let bgKnob = color('backgroundColor', Theme.colors.surface) // or whatever colors we want

useEffect(() => {
    bgKnob = color('backgroundColor', Theme.colors.surface) // or whatever color this should be
}, [darkMode]);
```

demos/storybook/package.json
daileytj: We're using this for the with-custom-menu UserMenu story, so I'm going to leave this in here for now unless we refactor that to use local state later.

demos/storybook/stories/drawer/with-basic-config.tsx
huayunh: I thought the idea is so that the users can copy from our code directly.

If these are shared then we need to include a comment that links to where this piece of code is on GitHub

joebochill: The previous implementation also used shared definitions of the menu items across stories. The primary use case of storybook is not to be able to copy and paste code into an application and have it work (unless they are building a storybook app) — it would be nice, but the way storybook works right now, I don't think it's possible (think about all of the knob code and the fact that the examples are not even react components, but `StoryFnReactReturnType`). We were hoping the storybook 6 story source plugin would give us more hope for being able to copy/paste from storybook, but that's still unknown.

The _design pattern_ examples' primary use case is for copy/pastabiity.

joebochill: I'm fine either way depending on what is more likely — are we more likely to want to make changes to the drawer items on a per-story basis (favor defining them individually) or are we more likely to want to make a change to the items that will be reflected in all of the stories (favor shared definition).

daileytj:
> > * The footer image swaps for the with-footer and full-config stories but there is a knob to change the background color so it's "hardcoded" when the theme is toggled. Should we still swap out the image even though we can't change the "hardcoded" color value? Or is there a way to make a knob default dynamic so we can hardcode the color with respect to the theme?
> 
> I haven't tried this before, but you could try using a `useEffect` hook to respond to a change to the `useDarkMode` value and then redeclare the knob with a new default value:
> 
> ```
> const darkMode = useDarkMode();
> let bgKnob = color('backgroundColor', Theme.colors.surface) // or whatever colors we want
> 
> useEffect(() => {
>     bgKnob = color('backgroundColor', Theme.colors.surface) // or whatever color this should be
> }, [darkMode]);
> ```

This ended up not working out the way I expected. There's a similar issue with the itemFontColor and itemIconColor when toggling the dark theme for the drawer full config story so I'm thinking we should just leave it as is for this story and open another issue to handle finding a solution for dynamic hard-coded values if we want to fix that.

daileytj:
I'm still getting a hooks issue when toggling to the rail variant on the within-a-drawer-layout story. Could someone else take a look at this?

joebochill:
> I'm still getting a hooks issue when toggling to the rail variant on the within-a-drawer-layout story. Could someone else take a look at this?

It's because you are calling a hook inside of a conditionally rendered element:
image

Move that hook call outside of the rendering (make a const var for it) and you should be good.

joebochill:
Let's log a new issue (or JIRA story) for looking into resetting the default knobs when swapping theme.

PR #282, opened 2021-05-11
Responsive UserMenu

daileytj:
Fixes PXBLUE-1829.

- Make the UserMenu responsive
- Add prettier config to the root so the changelog formats correctly


docs/UserMenu.md
```diff
@@ -55,17 +55,19 @@ const avatar = <Avatar><SendIcon/></Avatar>;
...
+| menu               | Custom MUI Menu displayed when Avatar is clicked                          | Material-UI `Menu` | no       |         |
+| menuGroups         | Groups of menu items that display                                         | `UserMenuGroups[]` | no       |         |
+| menuSubtitle       | Subtitle shown when menu is open                                          | `string`           | no       |         |
+| menuTitle          | Title shown when menu is open                                             | `string`           | no       |         |
+| MenuProps          | Property overrides for the MUI Menu                                       | `MenuProps`        | no       |         |
+| onClose            | Function called when the menu is closed                                   | `Function`         | no       |         |
+| onOpen             | Function called when the menu is opened                                   | `Function`         | no       |         |
+| useBottomSheetAt   | Window pixel width at which the responsive bottom sheet menu is triggered | `number`           | no       | 600     |
```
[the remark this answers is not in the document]
daileytj: true story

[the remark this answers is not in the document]
daileytj: > Not a disagreement, just curious what you think

I thought the same thing as I was adding the disableBottomSheet prop but figured it's better to be explicit about disabling things. I could go either way as long as we document it appropriately.

jeffvg:
looks good. Is it a preference to hide the footer in responsive view?

components/src/core/UserMenu/UserMenu.tsx
```diff
@@ -33,11 +33,19 @@ const useStyles = makeStyles(() =>
         noCursor: {
             cursor: 'inherit',
         },
+        bottomSheet: {
+            width: '100%',
+            maxWidth: 600,
+            margin: 'auto',
+            userSelect: 'none',
+            cursor: 'pointer',
```
joebochill, line 41: I'm not sure you want to assume clickable (pointer) for the entire bottomsheet. Probably better to let the individual menu items determine that, right?

components/src/core/UserMenu/UserMenu.tsx
```diff
@@ -200,7 +212,18 @@ const UserMenuRender: React.ForwardRefRenderFunction<unknown, UserMenuProps> = (
                 ...menu.props,
             });
         }
-        return (
+        return useMediaQuery(`(max-width:${useBottomSheetAt}px)`) && !disableBottomSheet ? (
+            <Drawer
+                data-cy="bottom-sheet"
+                anchor={'bottom'}
+                transitionDuration={250}
```
joebochill, line 219: This should probably use a transition duration value from the theme

components/src/core/UserMenu/UserMenu.tsx
```diff
@@ -200,7 +212,18 @@ const UserMenuRender: React.ForwardRefRenderFunction<unknown, UserMenuProps> = (
                 ...menu.props,
             });
         }
-        return (
+        return useMediaQuery(`(max-width:${useBottomSheetAt}px)`) && !disableBottomSheet ? (
```
joebochill, line 215: If it were me, I would declare this as a variable up above to help readability:
```
const showBottomSheet = useMediaQuery(...)
..
return showBottomSheet ? <BottomSheet> : <Menu>;
```

docs/UserMenu.md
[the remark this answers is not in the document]
joebochill: I would lean towards using the single prop...otherwise you've got two props that conflict with each other (i.e, if you didn't realize you had disabled on, you could be changing the usebotomsheetat to no avail).

You could cover this in the description, e.g.: '...menu is triggered (set to 0 to disable responsive behavior)'

package.json
```diff
@@ -41,6 +41,9 @@
     "author": "pxblue <pxblue@eaton.com>",
     "license": "BSD-3-Clause",
     "devDependencies": {
-        "codecov": "^3.8.1"
-    }
+        "@pxblue/prettier-config": "^1.0.3",
+        "codecov": "^3.8.1",
+        "prettier": "^2.3.0"
+    },
+    "prettier": "@pxblue/prettier-config"
```
joebochill, line 48: Does this do anything? There's no prettier script that runs at the root, it just jumps into the folders and runs it there I think:
image

docs/UserMenu.md
```diff
@@ -55,17 +55,19 @@ const avatar = <Avatar><SendIcon/></Avatar>;
...
 
 </div>
```
huayunh, line 72: I would suggest adding a note somewhere that says if you pass in a custom menu using the `menu` prop, then the responsiveness no longer applies.

components/src/core/UserMenu/UserMenu.tsx
```diff
@@ -200,7 +212,18 @@ const UserMenuRender: React.ForwardRefRenderFunction<unknown, UserMenuProps> = (
...
+            <Drawer
+                data-cy="bottom-sheet"
+                anchor={'bottom'}
+                transitionDuration={250}
+                open={Boolean(anchorEl)}
+                onClose={closeMenu}
+                classes={{ paper: clsx(defaultClasses.bottomSheet, classes.bottomSheet) }}
+            >
```
huayunh, line 223: Should we extend the `{...MenuProps}` here too?

huayunh: Or to simplify things a little bit, have some BottomsheetProp of the type DrawerProp

huayunh (commented):
- [ ] Add a screenshot in the readme that shows how this thing looks on mobile
- [ ] On mobile, the menu header has a height of `64px`/`4rem` in design
- [ ] Should we add a `useBottomSheetAt` knob in Storybook's full config story?

I can see how in the future the bottom sheet / responsive bottom sheet would allow your user to swipe down inside the sheet to dismiss it, in addition to tapping on the scrim.

components/src/core/UserMenu/UserMenu.tsx
```diff
@@ -70,6 +78,7 @@ const UserMenuRender: React.ForwardRefRenderFunction<unknown, UserMenuProps> = (
         MenuProps,
         menuSubtitle,
         menuTitle,
+        useBottomSheetAt = 600,
```
joebochill, line 81: `theme.breakpoints.values.sm` perhaps?

components/src/core/UserMenu/UserMenu.tsx
joebochill: Yeah we can do that...I would prefer the second option (`BottomsheetProps`) since the types for Menu and Drawer are not compatible.

I'm not sure if people are really going to need to change much with the drawer (unlike with the menu where they are likely to want to control it's position), but it doesn't hurt to give more customizability We've certainly given people enough options to do crazy things in the other components 😆

docs/UserMenu.md
joebochill: I would do that in the initial description paragraph (you can add a blip about how it switches between menu and bottomsheet, but if you supply your own menu you have to do it yourself).

components/src/core/UserMenu/UserMenu.tsx
```diff
@@ -33,11 +33,18 @@ const useStyles = makeStyles(() =>
         noCursor: {
             cursor: 'inherit',
         },
+        bottomSheet: {
+            width: '100%',
+            maxWidth: 600,
```
huayunh, line 38: If you set `useBottomSheetAt = theme.breakpoints.values.sm,` then I expect every "600" in your code to get changed

docs/UserMenu.md
```diff
@@ -55,17 +58,18 @@ const avatar = <Avatar><SendIcon/></Avatar>;
...
+| menuSubtitle     | Subtitle shown when menu is open                                                                                    | `string`           | no       |         |
+| menuTitle        | Title shown when menu is open                                                                                       | `string`           | no       |         |
+| MenuProps        | Property overrides for the MUI Menu                                                                                 | `MenuProps`        | no       |         |
+| onClose          | Function called when the menu is closed                                                                             | `Function`         | no       |         |
+| onOpen           | Function called when the menu is opened                                                                             | `Function`         | no       |         |
+| useBottomSheetAt | Window pixel width at which the responsive bottom sheet menu is triggered (set to 0 to disable responsive behavior) | `number`           | no       | 600     |
```
huayunh, line 72: This "600" shall get changed to `theme.breakpoints.values.sm` too

jeffvg (approved):

docs/UserMenu.md
```diff
@@ -1,12 +1,15 @@
 # User Menu
 
-The `<UserMenu>` is an Avatar that opens a Menu when clicked. It is typically used in the top-right corner of an application and indicates who is logged in.
-
-<div style="align-items: center; display:flex; justify-content: space-around">
+The `<UserMenu>` is an Avatar that opens a Menu when clicked. It is typically used in the top-right corner of an application and indicates who is logged in. By default, the Menu will responsively transition to a bottomsheet for mobile views.
+Passing in a custom menu using the `menu` prop or setting the `useBottomSheetAt` prop to zero will disable the responsiveness.
```
joebochill, line 4:
```suggestion
The `<UserMenu>` is an Avatar that opens a Menu when clicked. It is typically used in the top-right corner of an application and indicates who is logged in. By default, the Menu will responsively transition to a bottom sheet for mobile views (if passing in a custom menu, you will be responsible for handling any responsiveness on your content). Setting the `useBottomSheetAt` prop to zero will disable the responsiveness.
```

joebochill (approved):

joebochill:
> * [ ]  On mobile, the menu header has a height of `64px`/`4rem` in design

I feel like we should update the spec on this to be consistent with what we do for AppBars (64 desktop, 56 mobile). We're sharing the same component with the DrawerHeader here which is why it's 56.

PR #292, opened 2021-06-08
Basic Collapsible Toolbar (Round 2)

joebochill:
For PXBLUE-2111

- Implement a basic toolbar that collapsed when the page is scrolled past threshold
- Add docs
- Add basic tests

#### To Test:
In the showcase, in SharedAppBar.tsx replace the MUI AppBar with the AppBar imported from our component library.


docs/AppBar.md
```diff
@@ -0,0 +1,58 @@
+# App Bar
+
+The `<AppBar>` component is an extension of the default AppBar from Material UI that can be resized / collapsed as the page is scrolled. It supports three modes:
```
emclaug2, line 3: "it supports three modes:" is followed by 2 pictures and then the usage section below the pictures.  It's a little weird reading this because it's not explicit what the 3 modes are.

docs/AppBar.md
```diff
@@ -0,0 +1,58 @@
...
+| ----------------- | ------------------------------------------------ | -------------------------------------------- | -------- | ------------- |
+| animationDuration | Length of the collapse / expand animation (ms)   | `number`                                     | no       | theme default |
+| backgroundImage   | Image to use as the app bar background           | `string`                                     | no       |               |
+| classes           | Style Overrides                                  | `AppBarClasses`                              | no       |               |
+| collapsedHeight   | Height of the AppBar when collapsed              | `number`                                     | no       | theme default |
+| expandedHeight    | Height of the AppBar when expanded               | `number`                                     | no       | 200           |
+| mode              | Behavior mode of the App Bar                     | `'expanded'` \| `'collapsed'` \| `'dynamic'` | no       | 'dynamic'     |
+| scrollThreshold   | Distance to scroll before collapsing the app bar | `number`                                     | no       | 100           |
```
emclaug2, line 43: I updated my default scroll threshold to be `(expandedHeight-collapsedHeight)/2` since I thought that's what we talked about before.

components/src/core/AppBar/AppBar.tsx
```diff
@@ -0,0 +1,299 @@
...
+        // onExpandedHeightReached,
+        // onCollapsedHeightReached,
+        scrollThreshold = 100,
+        ...muiAppBarProps
+    } = props;
+    const theme = useTheme();
+    const defaultClasses = useStyles(props);
+    const animationDuration = durationProp || theme.transitions.duration.standard;
+    const isMobile = useMediaQuery(theme.breakpoints.down('xs'));
+    const defaultAppBarHeight = isMobile ? theme.spacing(7) : theme.spacing(8);
```
huayunh, line 143: If you want to go fancy about it, `theme.mixins.toolbar` is what you shall be looking into. https://material-ui.com/customization/default-theme/

emclaug2:
So this component doesn't support scroll listeners on elements instead of the window?  `scrollRef` should be moved over to a future feature?

docs/AppBar.md
huayunh, line 43: `scrollThreshold` default value shall prob be corrected. Same with your comments above the type definition.

docs/AppBar.md
```diff
@@ -0,0 +1,58 @@
...
+
+</div>
+
+Any other props supplied will be provided to the root element ([`AppBar`](https://material-ui.com/api/appbar/)).
```
huayunh, line 47: wrong MUI link

huayunh (approved):
Functionality wise looks good as a first pass

docs/AppBar.md
```diff
@@ -0,0 +1,58 @@
...
+### Classes
+
+You can override the classes used by PX Blue by passing a `classes` prop. It supports the following keys:
+
+| Name       | Description                                          |
+| ---------- | ---------------------------------------------------- |
+| root       | Styles applied to the root element                   |
+| background | Styles applied to the background image               |
+| expanded   | Styles applied to the root element when expanded     |
+| background | Styles applied to the background image when expanded |
```
huayunh, line 58: expandedBackground

docs/AppBar.md
joebochill: Yeah I'll reorganize this a bit....it doesn't look weird when you're just looking at the markdown file 😛

docs/AppBar.md
joebochill: I remember we were playing around with that, I didn't know that we settled on anything. The 'half-difference' feels like a really long way to scroll if the difference is large. My thinking is that the toolbar should collapse almost right away when you start scrolling (what's the point of keeping it big and limiting how much of the _real_ content a user sees?), but I'm open to discussion on it.

joebochill:
> So this component doesn't support scroll listeners on elements instead of the window? `scrollRef` should be moved over to a future feature?

Correct, for now I have left it out until a use case presents itself (the window element scrolling is the preferred behavior in most cases anyway for the mobile compatibility). I got it mostly working in a POC (code is still there just commented out), but there were some nuances that would need to be worked out.

docs/AppBar.md
joebochill: Oh, I definitely wrote descriptions at one point...I think they got lost on the day when I accidentally deleted the local file before saving it after moving it into the components library.

components/src/core/AppBar/AppBar.tsx
joebochill: `theme.mixins.toolbar['media (min-width:0px) and (orientation: landscape)'].minHeight` doesn't feel quite as clean...I presume it would also not work if somebody changes the default breakpoints in their theme. I'm actually going to tweak this to use REMS so it will exactly match what we're doing in the DrawerHeader

docs/AppBar.md
joebochill: I remember you made this same comment on the last PR...i've got it now

joebochill:
@huayunh I know you were also asking for a gif in the readme, but I left it super basic for now, figuring we would go in and add better visuals once it's fully done with the three-liner and so-on.

PR #293, opened 2021-06-17
Feature/three line component

emclaug2:
- Add 3 Liner component with some storybook examples


emclaug2:
Hosted here: https://pxblue-react-library.web.app/?path=/story/intro-overview--px-blue-react-components

huayunh:
That collapsed typography style doesn't look right to me.

components/src/core/AppBar/AppBar.tsx
```diff
@@ -16,6 +16,7 @@ export type AppBarClasses = {
     root?: string;
     background?: string;
     expanded?: string;
+    collapsed?: string;
```
joebochill, line 19: What is the difference between `collapsed` and `root`?

components/src/core/ThreeLiner/ThreeLiner.tsx
```diff
@@ -0,0 +1,94 @@
...
+    info?: string;
+};
+
+const useStyles = makeStyles((theme: Theme) =>
+    createStyles({
+        root: {
+            display: 'flex',
+            height: '100%',
+            flexDirection: 'column',
+            justifyContent: 'center',
```
joebochill, line 18: Just curious, does this need to be flex? It seems like it should work as block

components/src/core/ThreeLiner/ThreeLiner.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Custom classes to add to app bar elements
+     * Default: none
+     */
+    classes?: Partial<ThreeLinerClasses>;
+};
+
+const ThreeLinerRenderer: React.ForwardRefRenderFunction<unknown, ThreeLinerProps> = (props: ThreeLinerProps) => {
+    const { title, subtitle, info, classes = {}, className, ...otherDivProps } = props;
```
joebochill, line 82: you'll need to destructure `animationDuration` here (with an eslint ignore for unused vars) or it will get passed to the root div and throw a warning in the console.

components/src/core/ThreeLiner/ThreeLiner.tsx
```diff
@@ -0,0 +1,94 @@
...
+            fontWeight: 300,
+        },
+    })
+);
+
+export type ThreeLinerProps = HTMLAttributes<HTMLDivElement> & {
+    /**
+     * First Line Content
+     */
+    title?: string | JSX.Element;
```
joebochill, line 57: These should be `ReactNode`, which is 'anything that can be rendered' (numbers, boolean, string, JSX Element)

docs/ThreeLiner.md
```diff
@@ -0,0 +1,27 @@
+# Three Liner
+
+The `ThreeLiner` can display up to three lines of stylized text or other ng-content. It is most commonly used within the context of a [`AppBar`](./AppBar.md) component where the text can grow / shrink as the App Bar is expanded and collapsed.
```
joebochill, line 3: ng-content doesn't exist in react 😛

joebochill (commented):
Why are we deleting yarn.lock?

components/src/core/ThreeLiner/ThreeLiner.tsx
emclaug2: These are updated now to be ReactNodes

docs/ThreeLiner.md
emclaug2: Good catch, it's now "custom" content.

components/src/core/ThreeLiner/ThreeLiner.tsx
emclaug2: This has been destructured, should be good now

components/src/core/AppBar/AppBar.tsx
emclaug2: I added in the `collapsed` class because I only want these styles to be applied to a collapsed appbar.  If I put some of these styles on the root, I'd have to add additional styles to the `expanded` class set to disable the root styles that I no longer want.

components/src/core/ThreeLiner/ThreeLiner.tsx
emclaug2: I used flex here so that the three liner would be center aligned when the toolbar is collapsed.  This is a little weird because the positioning here does not match angular.

emclaug2:
> Why are we deleting yarn.lock?

This should not be deleted; I forgot to add this file!

emclaug2:
> That collapsed typography style doesn't look right to me.

I think the styles should be better now.  I'm going to redeploy the storybook app.

emclaug2:
Appbar hosted here: https://pxblue-react-library.web.app/?path=/story/*

demos/storybook/stories/three-liner/with-custom-content.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { ChannelValue, ThreeLiner } from '@pxblue/react-components';
+import Trend from '@material-ui/icons/TrendingUp';
+import * as Colors from '@pxblue/colors';
+import { getLeftToRightIconTransform } from '../../src/utils';
+
+export const withCustomContent = (): StoryFnReactReturnType => {
+    const title = text('title', 'title');
+    const subtitle = text('subtitle', 'subtitle');
+    const info = text('info', 'info');
+    const infoContent = select('infoContent', ['text', 'channelValue'], 'channelValue');
```
huayunh, line 13:
```suggestion
    const infoContent = select('infoContent', ['text', '<ChannelValue />'], '<ChannelValue />');
```

huayunh (approved):
Visually looks good!

Right now the Appbar is covering up the body content. I assume there would be a future polishment PR that allows the user to scroll to the top of Lorem Ipsum?

image

components/src/core/AppBar/AppBar.tsx
joebochill: You either have to override default styles from root, or add additional styles to collapsed. I think it's the same amount of work either way, but I agree with adding the collapsed class since it's much more obvious how you'd go about using it.

joebochill (approved):

demos/storybook/stories/app-bar/with-dynamic-content.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { makeStyles } from '@material-ui/core';
+
+export const withDynamicContent = (): StoryFnReactReturnType => {
+    const useStyles = makeStyles({
+        title: {},
+        subtitle: {},
+        info: {},
+        liner: {
+            top: 0,
+            marginLeft: 24,
```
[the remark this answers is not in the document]
emclaug2: Ok this is updated now; I didn't see any other storybook stories using theme.spacing, so I just used margin 32px in the meantime.

emclaug2:
> Right now the Appbar is covering up the body content. I assume there would be a future polishment PR that allows the user to scroll to the top of Lorem Ipsum?

I would mark this as a future enhancement because this is one of the first things I've tried to setup correctly.  Because we're in storybook, we don't have as much flexibility as we would if we were in a normal app.  This is a little tricky.

PR #332, opened 2021-11-11
Update changelog and package version

emclaug2:
-  Get ready to publish 5.3.2


emclaug2:
v5.3.2 has a beta package that is published and can be tested inside the showcase demo.

daileytj (approved):

PR #339, opened 2021-11-26
Add condition for rendering subtitle and info

ektasawant:
Fixes [318](https://github.com/pxblue/react-component-library/issues/318) .

- Add condition for adding typography to check subtitle prop has value
- Add condition for adding typography to check info prop has value

#### Screenshots:
Title + Subtitle
Screenshot 2021-11-26 at 4 08 55 PM

Only Title 
Screenshot 2021-11-26 at 4 11 00 PM

#### To Test:
- http://localhost:6006/?path=/story/components-info-list-item--with-subtitle

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
+## v5.3.3 (December 2, 2021)
 
+### Fixed
+
+-   Avoid rendering of HTML elements in DOM when `Subtitle` or `Info` prop is not available.
 ## v5.3.2 (November 11, 2021)
 
 ### Changed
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
+## v5.3.3 (December 2, 2021)
 
+### Fixed
+
+-   Avoid rendering of HTML elements in DOM when `Subtitle` or `Info` prop is not available.
```
emclaug2, line 6: Can you remove the capital letters from "Subtitle" & "Info" ?

emclaug2 (approved):
Looks good just see my comment about the docs; tested it in storybook & saw no element rendered in dom.

PR #343, opened 2021-12-15
Update documentation images

daileytj:
Fixes 1/3rd of PXBLUE-2656.

- Update documentation images so there is no reference to PX Blue

### Before
Screen Shot 2021-12-15 at 10 32 49 AM
Screen Shot 2021-12-15 at 10 32 56 AM
Screen Shot 2021-12-15 at 10 33 19 AM

### After
Screen Shot 2021-12-15 at 10 34 05 AM
Screen Shot 2021-12-15 at 10 34 14 AM
Screen Shot 2021-12-15 at 10 33 56 AM


emclaug2 (approved):
Looks great, once CI passes you're good to merge it.

huayunh:
I know this is only 1/3 of the story, but just a little heads up that our pics are obsolete style-wise. So like in this drawer, our default style for menu item is regular font weight (body 1) for an unselected menu item, and the icon colors shall be black500 (I know our API allows something crazier). I think the default style for nav group title is overline and black500.

image

For an anatomy page, we may want to show the default behavior to not confuse people — your call.

daileytj:
@huayunh I'd prefer to keep this story to "remove px blue references". If we want to update the style stuff that should be revisited in another story. I personally think all of these examples should be properly built out in Figma so they can be easily modified and updated when we make style changes. Some of them are built with Figma components already, but some are just old screenshots that aren't easily reproducible. For this specific example, I just overlayed a header component over the old screenshot.

daileytj:
@huayunh I'll wait for a response or approval before merging this just in case you still want changes made.

jeffvg (approved):

huayunh:
if you do not plan to fix the styling part this round, then you would want to make another story for it (or does it already exist?)

emclaug2:
> if you do not plan to fix the styling part this round, then you would want to make another story for it (or does it already exist?)

Yeah @daileytj can you make a follow-up story and throw it into R23?

daileytj:
> > if you do not plan to fix the styling part this round, then you would want to make another story for it (or does it already exist?)
> 
> @daileytj can you make a follow-up story and throw it into R23?

You got it!

PR #356, opened 2022-01-07
Make changelog more clear

huayunh:
- Make the changelog more readable

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 
 ### Fixed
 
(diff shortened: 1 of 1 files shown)


andrupu:
Another thought would be to word it like "Render HTML elements only when `subtitle` or `info` prop is available/specified in `<InfoListItem>`'s DOM tree."
Yeah, this is a tricky one to put into a sentence.  I am tongue twisted too.  I like how you included the tags / actual code.  I will leave it up to you to decide to go with your original re-write or my suggestion.  I like both descriptions, you have my approval either way.

andrupu (approved):

PR #360, opened 2022-01-12
Add 1rem padding to root of the EmptyState component

### LINKED ISSUE #320: Empty State shall have padding at the root DOM
Same as https://github.com/pxblue/angular-component-library/issues/336

daileytj:
Fixes #320.

- Add 1rem padding to root of the `EmptyState` component
- Removes 100% height and min-height styles from root of the `EmptyState` component

#### Screenshots:
Screen Shot 2022-01-12 at 2 36 46 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 -   Added `chevronColor` property onto `<InfoListItem>`.
 -   Added `chevronIcon` class for `chevron` element in `<InfoListItem>`.
+-   Added 1rem default padding to the root styles of `<EmptyState>`. ([#320](https://github.com/brightlayer-ui/react-component-library/issues/320))
 
 ## v5.3.3 (December 6, 2021)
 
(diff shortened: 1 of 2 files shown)


ektasawant:
Changes looks good 👍 . Is there any specific  reason to remove 100% height and min-height styles? I check few React design pattern example which uses empty state will behave like below images (after removing 100% height and min-height ).

URL: /in-button-panel
Screenshot 2022-01-13 at 3 59 52 PM

URL: /progress-bar-indeterminate
Screenshot 2022-01-13 at 4 04 46 PM

joebochill:
The 100% height styles are in there because the EmptyState component is supposed to fill whatever container it sits in (and then use flexbox to center the icon and text in the middle). Removing those styles will break everywhere the empty state is used currently (examples, workflows, etc.)

emclaug2 (approved):
Keeping height at 100%, we might need to double-check the angular implementation because I think it's not taking up 100% height of the parent container.

PR #361, opened 2022-01-13
Added titleDivider prop to DrawerNavGroup

### LINKED ISSUE #315: DrawerNavGroup Title Divider Prop
A user should be able to toggle the divider via a prop (e.g., `titleDivider: boolean`)

Add this prop to the storybook stories for 'full config' and 'with multiple nav groups'
If the DrawerNavGroup has a title, a divider is provided below the title by default.

#### Additional Context

We may want to have a `titleDividerPosition: 'top' | 'bottom'` prop as well, so that the divider can be positioned above or below the DrawerNavGroup title depending on the context.

bkarambe:
Fixes #315 

- Added titleDivider prop to DrawerNavGroup

#### Screenshots:
**Before**
Screenshot 2022-01-13 at 4 22 42 PM

**After**
Screenshot 2022-01-13 at 4 21 43 PM

#### To Test:
- yarn start:showcase
- Pass titleDivider={false} to DrawerNavGroup on Dashboard

### DIFF
--- a/components/src/core/Drawer/DrawerNavGroup.tsx
+++ b/components/src/core/Drawer/DrawerNavGroup.tsx
@@ -33,6 +33,9 @@ export type DrawerNavGroupProps = SharedStyleProps &
 
         /** Custom element, substitute for title */
         titleContent?: ReactNode;
+
+        /** Divider for the title */
+        titleDivider?: boolean;
     };
 
 type DrawerNavGroupClasses = {
@@ -106,6 +109,7 @@ const DrawerNavGroupRender: React.ForwardRefRenderFunction<unknown, DrawerNavGro
         title,
         titleColor = theme.palette.text.primary,
         titleContent,
+        titleDivider = true,
         // Shared Style Props
         activeItemBackgroundColor,
         activeItemBackgroundShape,
@@ -248,7 +252,9 @@ const DrawerNavGroupRender: React.ForwardRefRenderFunction<unknown, DrawerNavGro
                 }
                 {...otherListProps}
             >
-                {variant !== 'rail' && <div key={`${title}_title`}>{(title || titleContent) && <Divider />}</div>}
+                {variant !== 'rail' && (
+                    <div key={`${title}_title`}>{(title || titleContent) && titleDivider && <Divider />}</div>
+                )}
                 {items.map((drawerItem: DrawerNavItemProps | DrawerRailItemProps, index: number) => {
                     if (variant === 'rail') {
                         const railItem = drawerItem as DrawerRailItemProps;
--- a/components/src/core/Drawer/DrawerNavGroup.tsx
+++ b/components/src/core/Drawer/DrawerNavGroup.tsx
@@ -28,8 +28,14 @@ export type DrawerNavGroupProps = SharedStyleProps &
         /** Text to display in the group header */
         title?: string;
 
+        /** Color used for the title text */
+        titleColor?: string;
+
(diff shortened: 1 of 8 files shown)


daileytj:
I have not tested this yet, but you'll need to update storybook per the story details, `Add this prop to the storybook stories for 'full config' and 'with multiple nav groups'`

docs/Drawer.md
```diff
@@ -231,6 +231,7 @@ The `items` property supports nested items to generate collapsible sections in t
 | title                           | Text to display in the group header                        | `string`                | no       |                             |
 | titleColor                      | Color used for the title text                              | `string`                | no       | varies for light/dark theme |
 | titleContent                    | Custom element, substitute for title                       | `ReactNode`             | no       |                             |
+| titleDivider                    | Divider for the title                                      | `boolean`               | no       |
```
huayunh, line 234: Default is `true`

components/src/core/Drawer/DrawerNavGroup.tsx
```diff
@@ -33,6 +33,9 @@ export type DrawerNavGroupProps = SharedStyleProps &
 
         /** Custom element, substitute for title */
         titleContent?: ReactNode;
+
+        /** Divider for the title */
+        titleDivider?: boolean;
```
huayunh, line 38: You might want to add this as part of the `SharedStyleProps` instead in `/components/src/core/Drawer/types.tsx`, so that you can specify it at the top level of `<Drawer />`. 

The drawer structure is rather complicated, feel free to have a chat with any one of us.

huayunh: (That `titleColor` prop should also be part of the `SharedStyleProps`; not sure why it is listed here)

huayunh (commented):
Functionality-wise it is sound.

Also please address Tom's comment about the storybook.

joebochill (commented):
Like we discussed this morning, I'm not convinced we need to do the cascading prop inheritance for titleDivider since it only affects the DrawNavGroup (as you saw, we're not currently doing inheritance for the titleColor, which is similar). All of the other shared props are applicable to the DrawerNavItem at least. I could see an argument for ease of configuration by setting it on the Drawer and having it apply to all groups, so you might be able to convince me, but it can get pretty confusing setting something like `titleColor` on the drawer and it not affecting the `title` in the InfoListItem.

daileytj (approved):
Looks good! Let's get this thing merged in.

huayunh:
> @joebochill ... All of the other shared props are applicable to the DrawerNavItem at least.

We have two types, 
* `SharedStyleProps`, which controls everything from Drawer to DrawerBody to DrawerNavGroup, and 
* `NavItemSharedStyleProps`, which is like `SharedStyleProps` but goes all the way from Drawer to DrawerNavItem.

Was proposing adding `titleDivider` and `titleColor` to `SharedStyleProps`.

joebochill:
@huayunh that's not _completely_ accurate. `SharedStyleProps` and `NavItemSharedStyleProps` both propagate all the way to the `<DrawerNavItem>`. The reason why `NavItemSharedStyleProps` is separated is because those props ONLY apply to the `<DrawerNavItem>` and _not_ to the `<DrawerRAILItem>`. `SharedStyleProps` apply to both the NavItem and the RailItem. Like I said, I think a case could be made to add these new props, so let's have a chat about it and we can add another issue if we decide to go for it.

Issue #351, opened 2021-12-27
Add a ToolbarMenu component


joebochill:
#### Describe the desired behavior
This will be a replacement for the existing DropdownToolbar component. We will build a generic component that will handle the dropdown arrow iconography and state and trigger the menu open and closed. The user will supply a ReactNode for the label (could be text, could be icon + text, etc.) and we will append the dropdown arrow beside their content. Clicking the content will open the menu described by the user (either through an items prop, or custom menu). Most of this functionality can simply be taken out of the existing dropdown toolbar.

Add deprecation warnings for the DropdownToolbar. Make sure that instructions for using the new approach is mentioned in the readme for the Appbar component (e.g., as an example).

#### Describe the current behavior
Rather than having this functionality built into a Toolbar, we will isolate it into a smaller sub-component that can be used in any toolbar (regular Mui Appbar, blui collapsible, etc.)

#### Is this request related to a current issue?
No

ektasawant:
PR merge https://github.com/brightlayer-ui/react-component-library/pull/366. Closing this issue.

PR #366, opened 2022-01-24
Feature/2841 toolbar component

ektasawant:
Fixes [351](https://github.com/brightlayer-ui/react-component-library/issues/351) .

- [x] Add ToolbarMenu component
- [x] Update storybook
- [x] Add deprecation warning in dropdown toolbar component.

#### Screenshots:

Screenshot 2022-01-24 at 1 18 02 PM

Screenshot 2022-01-24 at 1 18 32 PM

Screenshot 2022-01-24 at 3 38 34 PM

#### To Test:
- yarn start:storybook from root
- http://localhost:6006/?path=/info/components-toolbar-menu--get-read-me-story


demos/storybook/stories/toolbar-menu/icon-with-label.tsx
```diff
@@ -0,0 +1,60 @@
...
+    const label = text('label', 'Subtitle');
+    const direction = getDirection();
+    return (
+        <ToolbarMenu
+            classes={{ label: classes.labelContent }}
+            label={
+                <span>
+                    <GradeA className={clsx(direction === 'rtl' ? classes.iconMarginLeft : classes.iconMarginRight)} />
+                    <span>{label}</span>
+                </span>
```
joebochill, line 53: If we're going to show a story for 'icon with label', in my mind, that would imply that we have a built-in functionality for including an icon with the label, versus combining the label and icon into the label prop. As an end-user I would be confused how this works if I was just looking at the story and not the source code for it.

docs/ToolbarMenu.md
```diff
@@ -0,0 +1,79 @@
+# Toolbar Menu
+
+The `ToolbarMenu` component is used to display a dropdown menu with label. The Menu can be populated via the `menuGroups` prop, or can be entirely customized by supplying your own `<Menu>` via the `menu` prop.
```
joebochill, line 3:
```suggestion
The `ToolbarMenu` component is used to display a dropdown menu with label, primarily inside of `<Toolbar>` or Card Header elements. The Menu can be populated via the `menuGroups` prop, or can be entirely customized by supplying your own `<Menu>` via the `menu` prop.
```

docs/ToolbarMenu.md
```diff
@@ -0,0 +1,79 @@
...
+          },
+          {
+            title: "Menu Item 3";
+            onClick: () => {};
+          }
+        ]
+      }
+    ]}
+/>
+```
```
joebochill, line 52: The first example in the Readme should be the most basic example possible with the minimum amount of excess imports or functionality:
```suggestion
```tsx
import { ToolbarMenu } from '@brightlayer-ui/react-components';
...
<ToolbarMenu
    label={'label'}
    menuGroups={[
      {
        items: [{
            title: "Menu Item 1";
            onClick: () => {};
          },
          {
            title: "Menu Item 2";
            onClick: () => {};
          },
        ]
      }
    ]}
/>
```
```

File: docs/ToolbarMenu.md [hunk 4]
```diff
@@ -0,0 +1,79 @@
...
+
+<div style="overflow: auto;">
+
+| Prop Name      | Description                                  | Type                  | Required | Default |
+| -------------- | -------------------------------------------- | --------------------- | -------- | ------- |
+| classes        | Style Overrides                              | `ToolbarMenuClasses`  | no       |         |
+| menu           | Custom content to be displayed in the menu   | Material-UI `Menu`    | no       |         |
+| menuGroups     | Groups of menu items to display              | `ToolbarMenuGroups[]` | no       |         |
+| MenuProps      | Property overrides for the MUI Menu          | `MenuProps`           | no       |         |
+| navigationIcon | Navigation Icon to be displayed              | `JSX.Element`         | no       |         |
```
joebochill, line 64: Is this a copy paste artifact?

docs/ToolbarMenu.md
```diff
@@ -0,0 +1,79 @@
...
+          }
+        ]
+      }
+    ]}
+/>
+```
+
+## API
+
+<div style="overflow: auto;">
```
joebochill, line 56: I think we usually do our prop and class tables in alphabetical order

joebochill (commented):
Code looks good.

Don't forget to add deprecation warnings for the DropdownToolbar component with instructions for migrating to using the ToolbarMenu instead.

demos/storybook/stories/toolbar-menu/icon-with-label.tsx
ektasawant: I will add a new prop `icon` in ToolbarMenu component and update PR.

docs/ToolbarMenu.md
ektasawant: Removed navigationIcon prop.

demos/storybook/stories/toolbar-menu/icon-with-label.tsx
ektasawant: I have added `color` and `icon` prop https://github.com/brightlayer-ui/react-component-library/pull/366/files#diff-0196ae54409f59dbbda8923dc349bc06e8a6e709ade5057a155fa9a7f7a2aa07R76 . Please review.

components/src/core/DropdownToolbar/DropdownToolbar.tsx
```diff
@@ -160,6 +160,12 @@ const DropdownToolbarRender: React.ForwardRefRenderFunction<unknown, DropdownToo
         [onOpen]
     );
 
+    useEffect(() => {
+        console.warn(
+            `DropdwonToolbar component has been deprecated and will be removed in version 6.0.0. Instead of DropdwonToolbar use ToolbarMenu component in regular Toolbar like Mui Appbar, BLUI Appbar.`
```
joebochill, line 166:
```suggestion
            `The DropdownToolbar component has been deprecated and will be removed in version 6.0.0. Instead of DropdownToolbar, use the ToolbarMenu component inside a MUI Toolbar or Brightlayer UI AppBar component.`
```

components/src/core/ToolbarMenu/ToolbarMenu.tsx
```diff
@@ -0,0 +1,281 @@
...
+import { Theme } from '@material-ui/core/styles/createMuiTheme';
+import useTheme from '@material-ui/core/styles/useTheme';
+import PropTypes from 'prop-types';
+
+export type ToolbarMenuClasses = {
+    root?: string;
+    dropdownArrow?: string;
+    icon?: string;
+    label?: string;
+    labelContent?: string;
```
joebochill, line 18: What's the difference between label and labelContent?

components/src/core/ToolbarMenu/ToolbarMenu.tsx
```diff
@@ -0,0 +1,281 @@
...
+    menuItem?: string;
+};
+
+export type ToolbarMenuItem1 = Omit<NavItem, 'itemID'> & { itemID?: string };
+export type ToolbarMenuGroup1 = {
```
joebochill, line 23: Can we remove the '1' from the end here?

components/src/core/ToolbarMenu/ToolbarMenu.tsx
```diff
@@ -0,0 +1,281 @@
...
+    classes?: Partial<ToolbarMenuClasses>;
+};
+
+const ToolbarMenuRenderer: React.ForwardRefRenderFunction<unknown, ToolbarMenuProps> = (
+    props: ToolbarMenuProps,
+    ref: any
+) => {
+    const {
+        classes = {},
+        /* eslint-disable @typescript-eslint/no-unused-vars */
```
joebochill, line 99: You need to re-enable this after the last expected unused variable. Right now, you've got that lint check turned off for the entire rest of the file.

If you want to disable it for a single line, use 'eslint-disable-next-line'

components/src/core/ToolbarMenu/ToolbarMenu.tsx
```diff
@@ -0,0 +1,281 @@
...
+        <span
+            ref={ref}
+            className={clsx(defaultClasses.root, classes.root, menuGroups || menu ? defaultClasses.cursorPointer : '')}
+            data-test={'wrapper'}
+            {...otherSpanProps}
+        >
+            <Typography
+                ref={anchor}
+                aria-haspopup="true"
+                className={clsx(defaultClasses.labelContent, classes.labelContent)}
```
joebochill, line 201: This gets a little confusing. When I think of label, I think of the text. I wouldn't expect the icon and the dropdown arrow to be grouped under 'labelContent'.

components/src/core/ToolbarMenu/ToolbarMenu.tsx
```diff
@@ -0,0 +1,281 @@
...
+                color={'inherit'}
+                component={'span'}
+                onClick={(): void => {
+                    openMenu(anchor.current);
+                }}
+            >
+                {icon && (
+                    <span className={clsx(defaultClasses.icon, classes.icon)} data-test={'icon'}>
+                        {icon}
+                    </span>
```
joebochill, line 211: This could possibly also throw some DOM validation errors, depending on what gets passed in. Since we're inside of a Span element, there might be limitations on what's allowed inside it (e.g., you can't put an H1 inside of a P tag...there might be similar restrictions for span). I'm wondering if we should move the icon and the dropdown arrow outside of the Typography element, and move the onClick handler to the root. I think we might be able to eliminate one of the wrappers here and simplify the class overrides.

huayunh (commented):
Would love to see the Toolbar Menu component in the "Three Liner with custom content" story too.

----

docs/images/ToolbarMenuAnatomy.png does not contain the `icon` prop

----
Perhaps you would want to adjust the line height / height / margins of the menu. 

This is the existing dropdown toolbar component:
image

This is the proposed Toolbar Menu
image

----
Our app bar appears off-centered on mobile
image

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 -   Added `titleDivider` property onto `<DrawerNavGroup>`. ([#315](https://github.com/brightlayer-ui/react-component-library/issues/315))
 -   Added 1rem default padding to the root styles of `<EmptyState>`. ([#320](https://github.com/brightlayer-ui/react-component-library/issues/320))
 -   Added new property `unitSpace` to `<ChannelValue>` to manage spacing between the value and units. ([#350](https://github.com/brightlayer-ui/react-component-library/issues/350))
+-   Added `<ToolbarMenu>` component. ([#351](https://github.com/brightlayer-ui/react-component-library/issues/351))
+-   Added Deprecation warnings for DropdownToolbar component that will be removing in version 6.0.0.
```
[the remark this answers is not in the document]
joebochill: I think it's okay to include...we did that for some of the similar items for RN component library when we did the deprecations.
```suggestion
-   Added Deprecation warnings for DropdownToolbar component that will be removed in version 6.0.0.
```

components/src/core/DropdownToolbar/DropdownToolbar.tsx
joebochill: Agree with Huayun that it would be nice to see a link to the ToolbarMenu docs page in here as well.

huayunh:
Somehow when I tried doing `yarn && yarn start` after a fresh clone, this showed up.

image

ektasawant:
> Somehow when I tried doing `yarn && yarn start` after a fresh clone, this showed up.
> 
> image

Need to run `yarn install` added package for multiple ref @seznam/compose-react-refs

huayunh:
are you going to update the react-demo-showcase and update the submodule pointer? `yarn && yarn start` is always expected to run in a fresh clone

components/src/core/ToolbarMenu/ToolbarMenu.tsx
```diff
@@ -0,0 +1,259 @@
...
+import { Theme } from '@material-ui/core/styles/createMuiTheme';
+import useTheme from '@material-ui/core/styles/useTheme';
+import Typography from '@material-ui/core/Typography';
+
+export type ToolbarMenuClasses = {
+    root?: string;
+    dropdownArrow?: string;
+    icon?: string;
+    label?: string;
+    menuItem?: string;
```
joebochill, line 19: Seems like there is some mismatching between the class overrides we are exposing for the end user and the classes that we are applying internally. I.e.,:

- we're exposing a `menuItem` class that never gets applied anywhere within the component (i think we can probably leave this out since they can override these through the item props already)
- we apply default classes to the autogenerated navgroups, but don't give users a class to override these styles

components/src/core/ToolbarMenu/ToolbarMenu.tsx
```diff
@@ -0,0 +1,259 @@
...
+                outline: 'none',
+            },
+        },
+        rotateDropdownArrow: {
+            transform: 'rotate(180deg)',
+        },
+    })
+);
+
+export type ToolbarMenuProps = HTMLAttributes<HTMLSpanElement> & {
```
joebochill, line 68: We should extend all of the TypographyProps here as well and spread them to the root element, so the user can do things like add `className` or `variant` props to change some of the appearance. Refer to the ListItemTag for how we did this there.

components/src/core/ToolbarMenu/ToolbarMenu.tsx
```diff
@@ -0,0 +1,259 @@
...
+                className={clsx(
+                    defaultClasses.root,
+                    classes.root,
+                    menuGroups || menu ? defaultClasses.cursorPointer : ''
+                )}
+                component={'span'}
+                data-test={'wrapper'}
+                onClick={(): void => {
+                    openMenu(anchor.current);
+                }}
```
joebochill, line 187: When you spread the Typography props in here, you'll need to be careful and order the props properly so none of the behavior gets broken:

```
<Typography
    // props that we set, but we don't mind / won't break if the user overrides (e.g., aria-haspopup)
    {...otherTypographyProps}
    // props that we will merge with what the user provides (e.g., className)
    // props that we set that we will not allow the user to override because it will break things (e.g., onClick)
/>
```

joebochill (commented):
Code looks good other than the one item about the classes and exposing the Typography props.

components/src/core/ToolbarMenu/ToolbarMenu.tsx
joebochill: I just tested out the toolbar menu in the showcase doing this:
image

And comparing it to the standard toolbar, it looks like this:
image


I could get the margin to look correct by using `classes={{root: classes.title}}` — it's just a little more inconvenient that using `className` (either one should work).

The bigger issue is that I cannot change the variant to be `h6` to match the expected style for a title line.

andrupu (commented):
It compiled for me and ran well for me.

ektasawant:
> Code looks good other than the one item about the classes and exposing the Typography props.

Added storybook for variant https://github.com/brightlayer-ui/react-component-library/pull/366/files#diff-76aece428f38cacb1b17616ac8c382b8c001bb277b8e73403d0b2046c99cfb65R43. @joebochill , Do I need to update showcase example for ToolbarMenu in this story?

jeffvg:
noticing the menu props example having issues with menu props transform vertical and RTL

http://localhost:6006/?path=/story/components-toolbar-menu--with-menu-placement-options

image

ektasawant:
> noticing the menu props example having issues with menu props transform vertical and RTL
> 
> http://localhost:6006/?path=/story/components-toolbar-menu--with-menu-placement-options
> 
> image
> 
> image

Hi @jeffvg , thanks for highlighting  rtl issue. I have updated code. Could you please check it?
Screenshot 2022-02-01 at 10 58 06 AM
Screenshot 2022-02-01 at 10 56 58 AM

jeffvg (approved):
looks good

components/src/core/ToolbarMenu/ToolbarMenu.tsx
```diff
@@ -0,0 +1,258 @@
...
+        <>
+            <Typography
+                ref={composeRefs(ref, anchor)}
+                aria-haspopup="true"
+                {...otherTypographyProps}
+                className={clsx(
+                    defaultClasses.root,
+                    classes.root,
+                    menuGroups || menu ? defaultClasses.cursorPointer : ''
+                )}
```
joebochill, line 182: Don't forget the className prop that can be passed in as well (otherwise we are overwriting it instead of merging)
```suggestion
                className={clsx(
                    defaultClasses.root,
                    classes.root,
                    props.className,
                    menuGroups || menu ? defaultClasses.cursorPointer : ''
                )}
```

components/src/core/ToolbarMenu/ToolbarMenu.tsx
```diff
@@ -0,0 +1,258 @@
...
+                outline: 'none',
+            },
+        },
+        rotateDropdownArrow: {
+            transform: 'rotate(180deg)',
+        },
+    })
+);
+
+export type ToolbarMenuProps = TypographyProps & {
```
joebochill, line 67: May want to omit the onClick here since we're overriding it and not merging with anything the user provides:
```suggestion
export type ToolbarMenuProps = Omit<TypographyProps, 'onClick'> & {
```

joebochill (commented):
Code changes look good. I didn't look closely at the demos or the docs, but hopefully everybody else did.

We do need to update the showcase, but we can make a separate issue for that.

Issue #377, opened 2022-02-21
Warning for TouchRippleProps in DrawerNavItemRender component (MUIv5)


ektasawant:
#### Describe the bug
Getting warning:
<img width="632" alt="Screenshot 2022-02-21 at 5 25 37 PM" src="https://user-images.githubusercontent.com/10433274/154952937-3c119ccc-65c4-4f28-b7bd-56b274798207.png">

#### What is the expected behavior?
Fix warning 
#### What are the steps to reproduce?
1. Go to /templates/dashboard in react showcase demo
2. Check console warning.

#### Environment

All	
#### Anything else to add?
In Muiv5, InfolistItem ripple feature was breaking because button prop has been removed from ListItemProps. To fix this we have added ListItemButton which will accept ripple prop. After adding ListItemButton component in InfolistItem. DrawerNavItem starts giving above warning.

ektasawant:
Fixed in https://github.com/brightlayer-ui/react-component-library/pull/380. Closing this issue.

PR #380, opened 2022-02-28
Feature/2926 refactor drawer nav item mui v5

ektasawant:
Fixes [377](https://github.com/brightlayer-ui/react-component-library/issues/377) .

- [x] Update drawerNavItem for using ListItemButtonProps
- [x] Update Drawer readme

#### Screenshots:
React Showcase Demo Drawer
Screenshot 2022-02-28 at 4 19 14 PM

InfolistItem (last example card in /templates/dashboard of showcase demo)
Screenshot 2022-02-28 at 4 20 24 PM

#### To Test:
1. Run yarn build from react-component-library 
2. Checkout branch name - feature/2711-upgrate-mui-to-v5.
3. Run yarn install in showcase demo and update node_modules/@brightlayer-ui/react-components with dist folder of react-component repository (Step 1)
4. Test drawer and Resources  example card from Local test URL: http://localhost:3000/templates/dashboard.

### DIFF
--- a/components/src/core/Drawer/DrawerHeader.tsx
+++ b/components/src/core/Drawer/DrawerHeader.tsx
@@ -213,6 +213,7 @@ const DrawerHeaderRender: React.ForwardRefRenderFunction<unknown, DrawerHeaderPr
                                     onIconClick();
                                 }}
                                 edge={'start'}
+                                size={'large'}
                             >
                                 {icon}
                             </IconButton>
--- a/components/src/core/Drawer/DrawerNavItem.tsx
+++ b/components/src/core/Drawer/DrawerNavItem.tsx
@@ -87,7 +87,7 @@ export type NestedNavItem = NestedDrawerNavItemProps;
 // First nested item has no additional indentation.  All items start with 16px indentation.
 const calcNestedPadding = (theme: Theme, depth: number): string => {
     const calculatePaddingInt = parseInt(theme.spacing(depth ? (depth - 1) * 4 : 0)) + parseInt(theme.spacing(2));
-    return `${calculatePaddingInt  }px`;
+    return `${calculatePaddingInt}px`;
 };
 
 const getChevronColor = (props: DrawerNavItemProps, theme: Theme): string => {
--- a/components/src/core/Drawer/DrawerNavItem.tsx
+++ b/components/src/core/Drawer/DrawerNavItem.tsx
@@ -24,6 +24,7 @@ export type DrawerNavItemClasses = {
     active?: string;
     expandIcon?: string;
     infoListItemRoot?: string;
+    listItemButtonRoot?: string;
     nestedListGroup?: string;
     nestedTitle?: string;
     title?: string;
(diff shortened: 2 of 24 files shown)


jeffvg:
we can verify this once showcase has been updated with RTL fix
image

components/src/core/Drawer/DrawerNavItem.tsx
```diff
@@ -403,7 +405,7 @@ const DrawerNavItemRender: React.ForwardRefRenderFunction<HTMLElement, DrawerNav
 
     // Combine the classes to pass down the the InfoListItem
     const infoListItemClasses = {
-        root: clsx(defaultClasses.infoListItemRoot, classes.infoListItemRoot),
+        root: ripple && hasAction ? undefined : clsx(defaultClasses.infoListItemRoot, classes.infoListItemRoot),
```
daileytj, line 408: Can you clarify why you are only applying these styles when there is no ripple and action?

components/src/core/Drawer/DrawerHeader.tsx
```diff
@@ -213,6 +213,7 @@ const DrawerHeaderRender: React.ForwardRefRenderFunction<unknown, DrawerHeaderPr
                                     onIconClick();
                                 }}
                                 edge={'start'}
+                                size={'large'}
```
ektasawant, line 216: We need `size={'large'}` which will match padding of our previous DrawerHeader.
size={'medium'} - default value
Screenshot 2022-03-02 at 4 13 06 PM

size={'large'}
Screenshot 2022-03-02 at 4 12 47 PM

jeffvg:
not sure if we talked about this or not. if so disregard this....

2711 branch missing ellipsis on top navs

image
 
before it was like this..
image

and user menu not themed yet

image

ektasawant:
> not sure if we talked about this or not. if so disregard this....
> 
> 2711 branch missing ellipsis on top navs
> 
> image
> 
> before it was like this.. image
> 
> and user menu not themed yet
> 
> image

In current alpha package of React-component drawer component is breaking. This is should be fix in this current PR. 
For testing this changes please follow below steps:

1. Run yarn build from root of react-component-library
2. Checkout to feature/2711-upgrate-mui-to-v5 in react-showcase-demo repo. And update node_modules/@brightlayer-ui/react-components with dist folder which is created by step 1.

> and user menu not themed yet
> 
> image
- This will be fix in https://github.com/brightlayer-ui/react-themes/pull/33

components/src/core/Drawer/DrawerNavItem.tsx
joebochill: When an InfoListItem has a click event associated with it, we have to render a ListItemButton inside it...this button needs to have the padding rather than the root element or the ripple will be inside of the padding.

components/src/core/Drawer/DrawerNavItem.tsx
```diff
@@ -134,6 +137,13 @@ const useStyles = makeStyles((theme: Theme) =>
             paddingRight: (props: DrawerNavItemProps): string =>
                 theme.direction === 'ltr' ? theme.spacing(2) : calcNestedPadding(theme, props.depth),
         },
+        listItemButtonRoot: {
```
joebochill, line 140: rather than defining the exact same rules twice, could we just have the one rule (`infoListItemRoot`) and conditionally apply it to either the InfoListItem or the ListItemButton as appropriate?

components/src/core/Drawer/DrawerNavItem.tsx
joebochill: However, I think we should still apply the user supplied classes here (classes.infoListItemRoot). Unless we go to having a single shared class as suggested above.

joebochill:
Code changes look good and it runs good. Looks nice with the alpha packages. I was able to simplify the testing procedure by simply dropping down into the showcase folder, checking out the feature branch you mentioned, and then running the `yarn start:showcase` from the root. No manual building or copying required.

joebochill:
Code changes look good and it runs good. Looks nice with the alpha packages. I was able to simplify the testing procedure by simply dropping down into the showcase folder, checking out the feature branch you mentioned, and then running the `yarn start:showcase` from the root. No manual building or copying required.

emclaug2:
> Code changes look good and it runs good. Looks nice with the alpha packages. I was able to simplify the testing procedure by simply dropping down into the showcase folder, checking out the feature branch you mentioned, and then running the `yarn start:showcase` from the root. No manual building or copying required.

Cool that's what I was doing as well, `yarn start:showcase` works great.

components/src/core/Drawer/DrawerNavItem.tsx
ektasawant: I have updated code and using shared class which will apply as per condition.

joebochill (approved):

PR #389, opened 2022-03-18
remove prop from redme

ektasawant:
Update readme.

### DIFF
--- a/docs/Hero.md
+++ b/docs/Hero.md
@@ -8,7 +8,43 @@ The Brightlayer UI `<Hero>` components are used to call attention to particular 
 
 The `<Hero>` component displays a particular icon, value/units, and a label. The icon property will accept any valid component - this will typically be a Material icon, [Brightlayer UI icon](https://github.com/brightlayer-ui/icons), or [Progress Icon](https://github.com/brightlayer-ui/icons/tree/master/progress). It will also accept Text/Emoji values.
 
-The value section of the Hero utilizes a [`<ChannelValue>`](./ChannelValue.md) component. To display a single simple value, the information can be passed as props (`value`, `units`, `valueIcon`). For more complex values (such as a duration that displays hours and minutes), you can pass in `<ChannelValue>` components as children and they will be displayed inline.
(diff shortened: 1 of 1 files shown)


docs/Hero.md
```diff
@@ -8,7 +8,43 @@ The Brightlayer UI `<Hero>` components are used to call attention to particular
 
 The `<Hero>` component displays a particular icon, value/units, and a label. The icon property will accept any valid component - this will typically be a Material icon, [Brightlayer UI icon](https://github.com/brightlayer-ui/icons), or [Progress Icon](https://github.com/brightlayer-ui/icons/tree/master/progress). It will also accept Text/Emoji values.
 
-The value section of the Hero utilizes a [`<ChannelValue>`](./ChannelValue.md) component. To display a single simple value, the information can be passed as props (`value`, `units`, `valueIcon`). For more complex values (such as a duration that displays hours and minutes), you can pass in `<ChannelValue>` components as children and they will be displayed inline.
+The value section of the Hero utilizes a [`<ChannelValue>`](./ChannelValue.md) component. To display a single simple value, the information can be passed using ChannelValueProps which will allow you to specify any props on the underlying `<ChannelValue>` component. `fontSize` prop has been removed from `<Hero>`. If you are using `fontSize` prop with value `small` replace it with following:
```
joebochill, line 11: This part about the fontSize prop should go in the changelog instead. The docs should always describe the current state of the component (rather than previous implementations).

And you can simplify the example in the changelog a bit with something like:

Changed:
- Removed `value`, `units`, `valueIcon`, and `fontSize` props — these should now be passed via the `ChannelValueProps` prop.
- Quick set options for `fontSize` ('normal' and 'small')  have been removed — if you were using these options previously, they can be replaced with '1.25rem' and '1rem', respectively (`<Hero ChannelValueProps={{fontSize: '1rem'}} />`).

joebochill: And in the hero, can you update:
- `values` class to include a style rule for fontSize: '1.25rem'
- remove the fontSize={20} on line ~116 (the automatically created ChannelValue element)

This will make the ChannelValue part of the Hero continue to have the correct size when no fontSize is provided in a declarative ChannelValue child (ChannelValue default is set to inherit, so the wrapper class values in the hero should set the size we want by default).

joebochill (commented):
Didn't we also need to change something in the package.json dependencies?

docs/Hero.md
ektasawant: Done. Please review .

jeffvg:
didn't look in to this yet but react-components 7.0.0-alpha.2 is causing unit tests to fail on drawer nav group and user menu
image

daileytj (approved):
Looks good! Great job

PR #394, opened 2022-04-07
Update ChannelValue to use Mui v5 styles

daileytj:
Fixes BLUI-3050.

- Update styles to use Mui-v5 Styled components
- Update component to extend `sx` prop
- Update component docs for `ChannelValue`

Test:
There is a showcase branch `feature/3050-channel-value-mui-v5-styles-update` that you can use to test this.
Also, test out the new `sx` prop extension. You can apply styles to the root by using `sx` directly on the component as well as by targeting nested items by the generated classnames. I will work on documentation while this is in draft review.

 ```tsx
<ChannelValue
     value={'123'}
     units={'hz'}
     color={'green'}
     classes={{ value: classes.value }}
     sx={{
            mt: 2,
            paddingLeft: '1rem',
            '& .BluiChannelValue-units': { pt: 4, backgroundColor: '#badabd', color: '#bada' },
            '& .BluiChannelValue-value': { pt: 4, backgroundColor: '#c00110', color: '#ee0e' },
      }}
/>
```

### DIFF
--- a/components/src/core/ChannelValue/ChannelValue.tsx
+++ b/components/src/core/ChannelValue/ChannelValue.tsx
@@ -7,7 +7,9 @@ import { styled } from '@mui/material/styles';
 import channelValueClasses, { ChannelValueClasses, getChannelValueUtilityClass } from './ChannelValueClasses';
 import { unstable_composeClasses as composeClasses } from '@mui/base';
 
-const useUtilityClasses = (ownerState: ChannelValueProps): any => {
+const useUtilityClasses = (
+    ownerState: ChannelValueProps
+): Record<'text' | 'prefix' | 'root' | 'icon' | 'suffix' | 'value' | 'units', string> => {
     const { classes } = ownerState;
 
     const slots = {
--- a/components/src/core/ChannelValue/ChannelValue.tsx
+++ b/components/src/core/ChannelValue/ChannelValue.tsx
@@ -116,7 +116,6 @@ const ChannelValueRender: React.ForwardRefRenderFunction<unknown, ChannelValuePr
         ...props,
     };
     const defaultClasses = useUtilityClasses(ownerState);
-    console.log('typeof default classes: ', typeof defaultClasses);
     const prefixUnitAllowSpaceList = ['$'];
     const suffixUnitAllowSpaceList = ['%', '℉', '°F', '℃', '°C', '°'];
 
--- a/components/src/core/ChannelValue/ChannelValue.tsx
+++ b/components/src/core/ChannelValue/ChannelValue.tsx
@@ -1,17 +1,29 @@
-import React, { useCallback, HTMLAttributes } from 'react';
+import React, { useCallback } from 'react';
 import Typography from '@mui/material/Typography';
-import makeStyles from '@mui/styles/makeStyles';
-import createStyles from '@mui/styles/createStyles';
-import clsx from 'clsx';
+import { cx } from '@emotion/css';
 import PropTypes from 'prop-types';
+import { Box, BoxProps } from '@mui/material';
+import { styled } from '@mui/material/styles';
+import channelValueClasses, { ChannelValueClasses, getChannelValueUtilityClass } from './ChannelValueClasses';
+import { unstable_composeClasses as composeClasses } from '@mui/base';
 
-export type ChannelValueClasses = {
-    root?: string;
-    icon?: string;
-    units?: string;
-    value?: string;
+const useUtilityClasses = (ownerState: any) => {
+    const { classes } = ownerState;
+
+    const slots = {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by ektasawant** [comment]
Screenshot 2022-04-08 at 12 40 43 PM

It is applying class css. But how can I couldn’t find sx props applied to channelValue. Must be I am missing something. Also in react-showcase-demo it is giving below error for me.
Screenshot 2022-04-08 at 12 42 35 PM

**Comment by daileytj** [comment]
@ektasawant can you show me the styles you were trying to apply via sx that weren't working.

**Comment by ektasawant** [comment]
> ```tsx
> sx={{
>            mt: 2,
>            paddingLeft: '1rem',
>            '& .BluiChannelValue-units': { pt: 4, backgroundColor: '#badabd', color: '#bada' },
>            '& .BluiChannelValue-value': { pt: 4, backgroundColor: '#c00110', color: '#ee0e' },
>      }}
> ```

The above example which is mentioned in PR. I am trying to check how it will apply to component.

**Comment by emclaug2** [comment]
I started testing this out in the showcase & it seems to be working without issues.  The observed style precedence is...

1. `SX styles` take precedence
2. User-provided `classes` prop
3. Internal component `defaultClasses` prop

Does this sound right?

**Comment by daileytj** [comment]
> I started testing this out in the showcase & it seems to be working without issues. The observed style precedence is...
> 
> 1. `SX styles` take precedence
> 2. User-provided `classes` prop
> 3. Internal component `defaultClasses` prop
> 
> Does this sound right?

True story and I just updated docs to state that sx overrides the classes prop

File: docs/ChannelValue.md [hunk 1]
```diff
@@ -43,9 +43,40 @@ Any other props supplied will be provided to the root element (`span`).
...
+### `sx` Class Overrides
+
+You can override the styles used by Brightlayer UI by passing a `sx` prop. The `sx` prop styles will override styles provided through the `Classes` prop. It supports the following classNames:
+
+| Selector                 | Description                          |
+| ------------------------ | ------------------------------------ |
+| .BluiChannelValue-icon   | Styles applied to the icon element   |
+| .BluiChannelValue-text   | Styles applied to the text element   |
+| .BluiChannelValue-prefix | Styles applied to the prefix element |
+| .BluiChannelValue-suffix | Styles applied to the suffix element |
```
huayunh, line 65: What is a prefix element? What is a suffix element?

daileytj: I get what you're getting at. There is a reference to what prefix is in the API and image at the top, but how would you suggest I describe what a prefix and suffix are?

daileytj: Would something such as "Styles applied to the unit element when used as a prefix|suffix" work as a description

huayunh: Looking at MUI's naming convention, I think the class names might make more sense this way:

`.BluiChannelValue-prefixUnit` `.BluiChannelValue-prefixValue`
`.BluiChannelValue-suffixUnit` `.BluiChannelValue-suffixValue`
(Yeah that's a lot)
or 
`.BluiChannelValue-prefixText` `.BluiChannelValue-suffixText`

If you just mention `.BluiChannelValue-prefix` it usually means "styles applied to the root element if `prefix={true}`"

You can also introduce an alias of `prefix` like `prefixed` to make the English flow better

Sorry just dumping all my thoughts. Or maybe you can choose not to advertise about it. I don't know man, these two classes feel strange

components/src/core/ChannelValue/ChannelValue.tsx
```diff
@@ -1,17 +1,29 @@
...
+import { styled } from '@mui/material/styles';
+import channelValueClasses, { ChannelValueClasses, getChannelValueUtilityClass } from './ChannelValueClasses';
+import { unstable_composeClasses as composeClasses } from '@mui/base';
 
-export type ChannelValueClasses = {
-    root?: string;
-    icon?: string;
-    units?: string;
-    value?: string;
+const useUtilityClasses = (ownerState: ChannelValueProps): any => {
```
bkarambe, line 10: Here type should be **_Record<"text" | "prefix" | "root" | "icon" | "suffix" | "value" | "units", string>_** instead of any

- [ ] Please replace type any

bkarambe (changes requested):

components/src/core/ChannelValue/ChannelValue.tsx
daileytj: good catch!

emclaug2 (approved):

daileytj:
@bkarambe Can you review this again? Your change request is blocking this right now.

bkarambe (approved):
Looks good.

PR #402, opened 2022-04-20
Updated EmptyState component to use mui v5 style engine

bkarambe:
Fixes BLUI-3053

Update component docs for EmptyState
Test:

<EmptyState
    sx={{margin: '0px'}}
    icon={<NotListedLocation />}
    title={'Location Unknown'}
/>

### DIFF
--- a/components/src/core/EmptyState/EmptyState.tsx
+++ b/components/src/core/EmptyState/EmptyState.tsx
@@ -1,19 +1,26 @@
-import React, { HTMLAttributes, ReactNode } from 'react';
-import makeStyles from '@mui/styles/makeStyles';
-import { useTheme, Theme } from '@mui/material/styles';
+import React, { ReactNode } from 'react';
+import { styled } from '@mui/material/styles';
+import { cx } from '@emotion/css';
 import Typography from '@mui/material/Typography';
-import clsx from 'clsx';
 import PropTypes from 'prop-types';
+import Box, { BoxProps } from '@mui/material/Box';
+import { EmptyStateClasses, EmptyStateClassKey, getEmptyStateUtilityClass } from './EmptyStateClasses';
+import { unstable_composeClasses as composeClasses } from '@mui/base';
 
-export type EmptyStateClasses = {
-    root?: string;
-    actions?: string;
-    description?: string;
-    icon?: string;
-    title?: string;
+const useUtilityClasses = (ownerState: EmptyStateProps): Record<EmptyStateClassKey, string> => {
+    const { classes } = ownerState;
+    const slots = {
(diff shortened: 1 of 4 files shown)


components/src/core/EmptyState/EmptyState.tsx
```diff
@@ -1,19 +1,26 @@
-import React, { HTMLAttributes, ReactNode } from 'react';
-import makeStyles from '@mui/styles/makeStyles';
-import { useTheme, Theme } from '@mui/material/styles';
+import React, { ReactNode } from 'react';
+import { styled } from '@mui/material/styles';
+import { cx } from '@emotion/css';
```
jeffvg, line 3: just a learning moment here but should this import be both cx and css?
import { cx, css } from '@emotion/css'

components/src/core/EmptyState/EmptyState.tsx
```diff
@@ -26,55 +33,72 @@ export type EmptyStateProps = HTMLAttributes<HTMLDivElement> & {
...
+})(({ theme }) => ({
+    color: theme.palette.mode === 'dark' ? theme.palette.text.secondary : theme.palette.text.primary,
+}));
+
+const Actions = styled(Box, {
+    name: 'empty-state',
+    slot: 'actions',
+})(() => ({
+    marginTop: '1rem',
 }));
```
daileytj, line 74: Let's not make all these styled-components. We can just style everything under the Root styled component. Look at how I did the default styles for ChannelValue.

components/src/core/EmptyState/EmptyState.tsx
```diff
@@ -26,55 +33,72 @@ export type EmptyStateProps = HTMLAttributes<HTMLDivElement> & {
...
     ref: any
 ) => {
-    const { actions, classes, description, icon, title, ...otherDivProps } = props;
-    const defaultClasses = useStyles(useTheme());
+    const { actions, classes, description, icon, title, ...otherProps } = props;
+    const ownerState = {
+        ...props,
+    };
+    const defaultClasses = useUtilityClasses(ownerState);
+
```
daileytj, line 84:
```suggestion
    const defaultClasses = useUtilityClasses(props);
```
we can pass props directly instead of storing it in this ownerState object

daileytj (commented):
This looks really good. Functionally everything is there, I just have some preferences for how we want to handle some things. 

There's also a little undocumented bug where we're not passing the classname. Can you follow what I did here and to pass the classname https://github.com/brightlayer-ui/react-component-library/pull/406

daileytj (approved):

PR #403, opened 2022-04-21
Feature/3054 update hero comp to use muiv5 style engine

bkarambe:
Fixes # BLUI-3054

Update component docs for Hero
Test:

<Hero sx={{ margin: '1rem' }} icon={<GradeA fontSize={'inherit'} />} label={'Efficiency'} style={{ flex: 1 }} />

### DIFF
--- a/components/src/core/Hero/Hero.tsx
+++ b/components/src/core/Hero/Hero.tsx
@@ -1,12 +1,12 @@
 import React, { ReactNode } from 'react';
 import Typography from '@mui/material/Typography';
 import { ChannelValue, ChannelValueProps as ChannelValuePropsType } from '../ChannelValue';
-import clsx from 'clsx';
+import { cx } from '@emotion/css';
 import PropTypes from 'prop-types';
 import Box, { BoxProps } from '@mui/material/Box';
 import { styled } from '@mui/material/styles';
 import { unstable_composeClasses as composeClasses } from '@mui/base';
-import { getHeroUtilityClass, HeroClasses, HeroClassKey } from './HeroClasses';
+import heroClasses, { getHeroUtilityClass, HeroClasses, HeroClassKey } from './HeroClasses';
 
 const useUtilityClasses = (ownerState: HeroProps): Record<HeroClassKey, string> => {
     const { classes } = ownerState;
@@ -43,7 +43,9 @@ export type HeroProps = BoxProps & {
     ChannelValueProps?: ChannelValuePropsType;
 };
 
-const Root = styled(Box, { name: 'hero', slot: 'root' })<Pick<HeroProps, 'onClick'>>(({ onClick, theme }) => ({
+const Root = styled(Box, { name: 'hero', slot: 'root' })<
+    Pick<HeroProps, 'onClick' | 'iconSize' | 'iconBackgroundColor'>
+>(({ onClick, iconSize, iconBackgroundColor, theme }) => ({
     fontSize: '1rem',
     display: 'flex',
     flexDirection: 'column',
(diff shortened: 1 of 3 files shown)


components/src/core/Hero/Hero.tsx
```diff
@@ -37,66 +43,74 @@ export type HeroProps = HTMLAttributes<HTMLDivElement> & {
...
+    slot: 'values',
+})(({ theme }) => ({
+    display: 'flex',
+    alignItems: 'center',
+    color: theme.palette.text.primary,
+    lineHeight: 1.2,
+    maxWidth: '100%',
+    overflow: 'hidden',
+    fontSize: '1.25rem',
+}));
```
daileytj, line 92: Let's not make all these styled-components. We can just style everything under the Root styled component. Look at how I did the default styles for ChannelValue.

components/src/core/Hero/Hero.tsx
```diff
@@ -37,66 +43,74 @@ export type HeroProps = HTMLAttributes<HTMLDivElement> & {
...
+    textOverflow: 'ellipsis',
+    whiteSpace: 'nowrap',
+}));
 
 const HeroRender: React.ForwardRefRenderFunction<unknown, HeroProps> = (props: HeroProps, ref: any) => {
-    const defaultClasses = useStyles(props);
+    const ownerState = {
+        ...props,
+    };
+    const defaultClasses = useUtilityClasses(ownerState);
```
daileytj, line 113: we can pass props directly instead of storing it in this ownerState object
```suggestion

    const defaultClasses = useUtilityClasses(props);
```

docs/Hero.md
```diff
@@ -102,3 +102,14 @@ You can override the classes used by Brightlayer UI by passing a `classes` prop.
 | Name | Description                        |
 | ---- | ---------------------------------- |
 | root | Styles applied to the root element |
+
+### `sx` Class Overrides
+
+You can override the styles used by Brightlayer UI by passing a `sx` prop. The `sx` prop styles will override styles provided through the `Classes` prop.
+
+```tsx
+import * as colors from '@brightlayer-ui/colors';
```
daileytj, line 111: Why is this in the example?

docs/Hero.md
```diff
@@ -102,3 +102,14 @@ You can override the classes used by Brightlayer UI by passing a `classes` prop.
 | Name | Description                        |
 | ---- | ---------------------------------- |
 | root | Styles applied to the root element |
```
daileytj, line 104: We should add the rest of the `classes` to the docs

daileytj:
The icon styles aren't working as they were before.

Screen Shot 2022-04-22 at 10 16 42 AM

emclaug2:
image

Yep just noticed this; icon styles again.

emclaug2:
So yesterday I was testing out the nested style rules can came across this observation:

This code snippet applies a red font onto a Hero's icon & works in the pre-MUI5 style migration:

```
<Hero
    classes={{
        icon: classes.red
    }}
    icon={
        <GradeA />
    }
    label={'Efficiency'} />
```

However after moving over to MUI5 styles, these overrides are ignored since the MUI5 styles introduce an additional layer of specificity to their style definitions.  The font isn't red anymore & the style hierarchy now looks like this:

image

Moving the icon's style override directly onto the component fixes the issue:

```
icon={<GradeA className={classes.red} fontSize={'inherit'} />
```

Unfortunately this problem is not only applied to the `Hero` component; it seems user-provided style-overrides via `classes` prop will be ignored within all non-root elements.

emclaug2:
After talking with Joe I'll make some follow-up stories to revisit the ChannelValue, Hero, EmptyState component so we can make sure these nested overrides still work via the `classes` prop.

emclaug2 (approved):

docs/Hero.md
```diff
@@ -99,17 +99,19 @@ Any other props supplied will be provided to the root element (`div`).
...
 
-| Name | Description                        |
-| ---- | ---------------------------------- |
-| root | Styles applied to the root element |
+| Name   | Description                          |
+| ------ | ------------------------------------ |
+| root   | Styles applied to the root element   |
+| icon   | Styles applied to the icon element   |
+| label  | Styles applied to the label element  |
+| values | Styles applied to the values element |
```
huayunh: why is there a mix of plural and singular

daileytj: I see your point, but this was in there prior to the changes made in this PR. This is out of scope for this issue and changing it now would be a breaking change. Please write a story to fix this in a separate milestone if you believe it to be an issue.

daileytj: The hero component can also take in multiple children, which would make the plural make sense since its styling all "values" whereas there will never be more than one label or icon.

daileytj (approved):

PR #405, opened 2022-04-21
Feature/3057 list item tag mui5 styles

emclaug2:
- MUI5 Style Engine for ListItemTag

#### To Test:
- Observed style override precedence for root overrides should be:
1. Style
2. ClassName
3. Classes
4. SX


daileytj (approved):

daileytj:
@bkarambe good catch on that base branch

PR #407, opened 2022-04-22
Feature/3055 hero banner use mui v5 style engine

bkarambe:
Fixes # BLUI-3055

Test:


daileytj:
This looks good @bkarambe. Can you just add the `sx` class overrides document updates to the herobanner component section in the Hero.md docs?

daileytj (approved):
Ignore my previous comment. I'm going to approve and merge this and write up a story to revisit all documentation at the end as we probably want to consolidate some of the repeated documentation around styling.

PR #409, opened 2022-04-26
Feature/3028 appbar mui5 styles

emclaug2:
- AppBar now MUI5 style engine


daileytj:
Can you update the submodule pointer to point back to dev?

daileytj:
@emclaug2 I hope you don't mind my pushing directly to your branch. It was just the className pass-through prop missing so className style overrides weren't working.

daileytj (approved):

daileytj:
I know I told you to point this at dev, but you end up needing to install @emotion/css. We may want to make a showcase update adding @emotion/css as a dependency that we can point at.

emclaug2:
> I know I told you to point this at dev, but you end up needing to install @emotion/css. We may want to make a showcase update adding @emotion/css as a dependency that we can point at.

I've opened up another PR for showcase demo; after that gets merged in we can update the pointer here again to fix the demo.

PR #410, opened 2022-04-27
Feature/3056 update infolistitem to use style engine

bkarambe:
Fixes # BLUI-3056

Test:

### DIFF
--- a/components/package.json
+++ b/components/package.json
@@ -14,6 +14,7 @@
     "prettier": "@brightlayer-ui/prettier-config",
     "dependencies": {
         "@brightlayer-ui/colors": "^3.0.0",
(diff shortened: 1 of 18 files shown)


emclaug2:
@bkarambe Can you reference some of the PRs I've done recently that wraps all non-root elements that require a `classes` override?

bkarambe:
> @bkarambe Can you reference some of the PRs I've done recently that wraps all non-root elements that require a `classes` override?

Sure. I will update the PR accordingly

bkarambe:
Closing this PR, as this will be handled in https://github.com/brightlayer-ui/react-component-library/pull/417

PR #411, opened 2022-04-27
Wrap each nested classes element

emclaug2:
- Each non-root element within ChannelValue that supports a user-provided class is now wrapped with `styled` .


daileytj (approved):
Code changes look good

PR #412, opened 2022-04-27
Wrap nested elements

emclaug2:
- Revisit Threeliner and wrap nested elements with `styled`


daileytj:
Can you use the `Box` component in place of `div` for the styled components.

daileytj (approved):
Code changes look good to me.

emclaug2:
> do we have to do anything with all this `webkit` stuff that's not being used?
> 
> image

I don't think so; all those crossed-off styles seem to have duplicated equivalents applied already.

PR #417, opened 2022-04-28
Feature/3056 update infolistitem comp mui v5 style engine

bkarambe:
Fixes # BLUI-3056

Test:


components/src/core/InfoListItem/InfoListItem.tsx
```diff
@@ -124,50 +155,64 @@ const InfoListItemRender: React.ForwardRefRenderFunction<unknown, InfoListItemPr
...
+                        {icon}
+                    </Icon>
                 </ListItemAvatar>
             );
         } else if (!hidePadding) {
             return (
                 // a dummy component to maintain the padding
                 <ListItemAvatar style={{ minWidth: 'unset' }}>
-                    <Avatar className={clsx(defaultClasses.avatar, defaultClasses.invisible)} />
+                    <InvisibleIcon
```
daileytj, line 185: Is this really an invisible icon? I would come up with a better way to name and organize this. This should probably be a styled avatar with the opacity style added to get the "invisible" effect.

emclaug2:
@bkarambe 

@daileytj made a PR to merge some changes into this branch, however this branch is now a little out of date with the `release/mui-v5-style-updates` branch.   Could you merge that release branch back into this feature branch so that Tom's PR is easier to understand?  I think Tom might have branched from the release branch instead of your feature branch.

jeffvg:
all the xs class overrides seem to be working except for BluiInfoListItem-icon. Not sure but wasn't able to give it any style override with the common sx props but could be me....

emclaug2:
@bkarambe  Can you open the console while running the showcase and observe these errors?

```
react-dom.development.js:67 Warning: React does not recognize the `statusColor` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `statuscolor` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```
We've seen something similar before; this happens when invalid props aren't being "de-spread" (I forget the jargon) from the props object before being passed onto a wrapped element.

components/src/core/InfoListItem/InfoListItemStyledComponents.tsx
```diff
@@ -0,0 +1,220 @@
...
+    position: 'absolute',
+    bottom: 0,
+    right: theme.direction === 'rtl' ? (divider === 'full' ? 0 : `4.5rem`) : 0,
+    left: theme.direction === 'ltr' ? (divider === 'full' ? 0 : `4.5rem`) : 0,
+    zIndex: 0,
+}));
+
+export const Icon = styled(Avatar, {
+    name: 'info-list-item',
+    slot: 'avatar',
```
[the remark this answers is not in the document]
bkarambe: I added 2 separate styled components Icon and IconAvatar

docs/InfoListItem.md
```diff
@@ -101,3 +101,39 @@ import { Link } from 'react-router-dom';
     button
 />
 ```
+
+### `sx` Class Overrides
```
daileytj, line 105: The formatting for this table needs fixed
Screen Shot 2022-05-16 at 9 13 43 AM

daileytj:
`sx={{ '& .BluiInfoListItem-listItemButtonRoot': { backgroundColor: '#bada55' } }}` This style is only applied if there is a ripple and a click event. Should this be applied to the listItemButtonRoot regardless of whether the ripple prop is on?

bkarambe:
> listItemButtonRoot
An element with className **listItemButtonRoot** only appears when ripple is on and there is onClick event.
This is code from base branch:
```
props.onClick && ripple ? (
                <ListItemButton className={combine('listItemButtonRoot')} focusRipple={ripple}>
                    {getInfloListItemContent}
                </ListItemButton>
```

daileytj (approved):

PR #432, opened 2022-05-11
Updated DrawerRailItem to work with mui v5 style engine

bkarambe:
Fixes # BLUI-3122

Test:
There is a showcase branch feature/3122-drawer-rail-item-example-showcase that you can use to test this.

### DIFF
--- a/components/src/core/Drawer/DrawerRailItem/DrawerRailItem.tsx
+++ b/components/src/core/Drawer/DrawerRailItem/DrawerRailItem.tsx
@@ -184,6 +184,7 @@ const StatusStripe = styled(Box, {
 const Icon = styled(Avatar, {
     name: 'drawer-rail-item',
     slot: 'icon',
+    shouldForwardProp: (prop) => prop !== 'itemIconColor',
 })<Pick<DrawerRailItemProps, 'itemIconColor'>>(({ itemIconColor, theme }) => ({
     color: itemIconColor || theme.palette.text.primary,
     backgroundColor: 'transparent',
--- a/components/src/core/Drawer/DrawerRailItem/DrawerRailItem.tsx
+++ b/components/src/core/Drawer/DrawerRailItem/DrawerRailItem.tsx
@@ -88,7 +88,11 @@ export type DrawerRailItemProps = SharedStyleProps & {
 const Root = styled(ButtonBase, {
     name: 'drawer-rail-item',
     slot: 'root',
-    shouldForwardProp: (prop) => prop !== 'backgroundColor',
+    shouldForwardProp: (prop) =>
+        prop !== 'backgroundColor' &&
+        prop !== 'activeItemIconColor' &&
+        prop !== 'activeItemFontColor' &&
+        prop !== 'statusColor',
 })<
     Pick<
         DrawerRailItemProps,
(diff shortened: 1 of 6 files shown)


emclaug2:
Taking a look at this now.

components/src/core/Drawer/DrawerRailItem/DrawerRailItem.tsx
```diff
@@ -68,109 +82,137 @@ export type DrawerRailItemProps = SharedStyleProps & {
...
+
+const DrawerRailItemDivider = styled(Divider, {
+    name: 'drawer-rail-item',
+    slot: 'divider',
+})(() => ({
+    position: 'absolute',
+    bottom: 0,
+    left: 0,
+    width: '100%',
+    [`& .${drawerRailItemClasses.itemActive}`]: {
```
emclaug2, line 211: Did you change this class name from `titleActive` to `itemActive` ?

components/src/core/Drawer/DrawerRailItem/DrawerRailItem.tsx
```diff
@@ -68,109 +82,137 @@ export type DrawerRailItemProps = SharedStyleProps & {
...
-            lineHeight: '1rem',
-            wordBreak: 'break-word',
-            hyphens: 'auto',
-            zIndex: 200,
-            color: (props): string => props.itemFontColor || theme.palette.text.primary,
+        [`& .${drawerRailItemClasses.condensed}`]: {
+            width: RAIL_WIDTH_CONDENSED,
+            minHeight: RAIL_WIDTH_CONDENSED,
         },
-        titleActive: {
```
emclaug2, line 168: Where did this class go?

components/src/core/Drawer/DrawerRailItem/DrawerRailItem.tsx
bkarambe: No. There are 2 separate classes titleActive and itemActive.

components/src/core/Drawer/DrawerRailItem/DrawerRailItem.tsx
bkarambe: titleActive class has been added to Title styled component

bkarambe:
> @bkarambe How did you go about testing these style changes?
> 
> Here's what I see in the showcase:
> 
> image
> 
> No SX classes are being applied. Am expecting `BluiDrawerRailItem-X` to be visible in the chain.
1

> @bkarambe How did you go about testing these style changes?
> 
> Here's what I see in the showcase:
> 
> image
> 
> No SX classes are being applied. Am expecting `BluiDrawerRailItem-X` to be visible in the chain.

There is no direct example to test this comp in the showcase. I have added a branch feature/3122-drawer-rail-item-example-showcase for this example.

bkarambe:
@emclaug2 After checking out thw showcase branch:
Screenshot 2022-05-16 at 2 33 21 PM

jeffvg:
trying to run showcase feature/3122-drawer-rail-item-example-showcase and getting this error...

image

code looks like this...does the sx just need moved elsewhere?

image

daileytj:
> trying to run showcase feature/3122-drawer-rail-item-example-showcase and getting this error...
> 
> image
> 
> code looks like this...does the sx just need moved elsewhere?
> 
> image

Did you try a fresh clone. Evan and I were having the same issue. Works fine on a fresh clone.

daileytj (approved):

PR #434, opened 2022-05-12
Updated Drawer comp to work with mui v5 style engine and updated docs

bkarambe:
Fixes #BLUI-3121

Test:


emclaug2:
> DrawerFooter

I'm guessing this PR is for the top-level Drawer and not DrawerFooter right?

bkarambe:
> > DrawerFooter
> 
> I'm guessing this PR is for the top-level Drawer and not DrawerFooter right?

Yes it is for top level Drawer comp

jeffvg (approved):
seems ok to me. Just need to update the doc for .BluiDrawer-expanded @emclaug2

daileytj:
sideBorder default styles and sx overrides are broken. Here's the result of adding these styles
`sideBorder={true}
  sx={{ '& .BluiDrawer-sideBorder': { width: '40px', backgroundColor: 'orange' } }}`


Screen Shot 2022-05-18 at 11 38 31 AM

jeffvg:
> sideBorder default styles and sx overrides are broken. Here's the result of adding these styles `sideBorder={true} sx={{ '& .BluiDrawer-sideBorder': { width: '40px', backgroundColor: 'orange' } }}`
> 
> Screen Shot 2022-05-18 at 11 38 31 AM

strang that looks diff than what I saw yesterday, I was using something like '& .BluiDrawer-sideBorder': { p: '24px', backgroundColor: 'green' } and had this...

image

daileytj:
hmmm... interesting @jeffvg. I'll try a fresh clone one more time just to be sure.

daileytj:
I tried on a fresh clone multiple times now with no luck on resolving the issue I have.

emclaug2:
@daileytj Here's what I see using your code snippet from above: 
image

bkarambe:
> sx={{ '& .BluiDrawer-sideBorder': { width: '40px', backgroundColor: 'orange' } }}

I tried by checking out dev branches on both comp lib and showcase. Applied these changes, but found the same result :
`sideBorder={true}
 classes={{sideBorder: classes.sideborder}}`

```
   sideborder: {
        width: '40px',
        backgroundColor: 'orange'
    }
```

Here's the screenshot after applying these changes :
MicrosoftTeams-image (5)

daileytj (approved):

PR #456, opened 2022-06-16
Publish v6.1.0

daileytj:
Publishes v6.1.0


jeffvg:
we test this with showcase - release/mui-v5-style-updates?

daileytj:
> we test this with showcase - release/mui-v5-style-updates?

That's one way to do it. Theres a beta package published though so you can test it in any app that is using our components.

ektasawant:
Using release/mui-v5-style-updates for testing all components. All components ate working fine. 

Note: Slowness issue in react showcase demo is still persist. Mostly when you click on input nav item.

daileytj:
@joebochill sounds good. I would propose that we don't use the custom slot name whatsoever for our internal styled functions. I don't see the need for them in any capacity.

components/src/core/Drawer/DrawerNavGroup/DrawerNavGroup.tsx
```diff
@@ -290,6 +303,8 @@ const DrawerNavGroupRender: React.ForwardRefRenderFunction<unknown, DrawerNavGro
                                 statusColor={railItem.statusColor}
                                 title={railItem.title}
                                 disableRailTooltip={railItem.disableRailTooltip}
+                                sx={railItem.sx}
+                                style={railItem.style}
```
joebochill, line 307: 297-307: it's strange that we are calling out all of these properties individually (error prone and not very future proof). Can't we just take those out and uncomment lioe 284 to spread all the props?

components/src/core/InfoListItem/InfoListItem.tsx
```diff
@@ -128,46 +158,73 @@ const InfoListItemRender: React.ForwardRefRenderFunction<unknown, InfoListItemPr
...
+                        <Icon
+                            statusColor={statusColor}
+                            iconColor={iconColor}
+                            avatar={avatar}
+                            iconAlign={iconAlign}
+                            className={combine('icon')}
+                        >
+                            {icon}
+                        </Icon>
+                    )}
```
joebochill, line 189: This is a lot of bloat...can we just put the conditional on the className prop instead since that's the only thing that's different between these two elements?

components/src/core/InfoListItem/InfoListItem.tsx
```diff
@@ -200,39 +257,50 @@ const InfoListItemRender: React.ForwardRefRenderFunction<unknown, InfoListItemPr
...
                         </>
                     ) : undefined
                 }
                 primaryTypographyProps={{
                     noWrap: !wrapTitle,
                     variant: 'body1',
+                    fontWeight: 600,
+                    lineHeight: 1.25,
+                    display: 'block',
+                    color: fontColor || 'inherit',
```
joebochill, line 303: What's the motive for specifying these values through props instead of CSS?

components/src/core/ListItemTag/ListItemTag.tsx
```diff
@@ -24,39 +35,41 @@ export type ListItemTagProps = TypographyProps & {
...
+// declare module '@mui/styles/withStyles' {
+//     // Augment the BaseCSSProperties so that we can control jss-rtl
+//     // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+//     interface BaseCSSProperties {
+//         /**
+//          * Used to control if the rule-set should be affected by rtl transformation
+//          */
+//         flip?: boolean;
+//     }
+// }
```
joebochill, line 47: If this is no longer needed can we remove it? How does MUI 5 style engine handle RTL style flipping?

docs/AppBar.md
```diff
@@ -68,14 +68,22 @@ import { AppBar, ThreeLiner } from '@brightlayer-ui/react-components';
...
+-   passing a `classes` prop with keys from the `Name` column below
+-   using the `Global CSS Class` in your main stylesheet
+
+For more details on styling options check out our [Styling Guide](https://github.com/brightlayer-ui/react-component-library/tree/master/docs#style-guide).
+
+| Name               | Global CSS Class               | Description                                          |
+| ------------------ | ------------------------------ | ---------------------------------------------------- |
+| root               | .BluiAppBar-root               | Styles applied to the root element                   |
+| background         | .BluiAppBar-background         | Styles applied to the background image               |
+| expanded           | N/A\*                         | Styles applied to the root element when expanded     |
```
joebochill, line 85: I don't get why this is NA. Shouldn't it just say '.Mui-expanded'?

docs/Drawer.md
```diff
@@ -90,17 +90,33 @@ The `Drawer` has four `variant`s:
...
+-   passing a `classes` prop with keys from the `Name` column below
+-   using the `Global CSS Class` in your main stylesheet
+
+For more details on styling options check out our [Styling Guide](https://github.com/brightlayer-ui/react-component-library/tree/master/docs#style-guide).
+
+| Name       | Global CSS Class       | Description                                                    |
+| ---------- | ---------------------- | -------------------------------------------------------------- |
+| root       | .BluiDrawer-root       | Styles applied to the root element                             |
+| content    | .BluiDrawer-content    | Styles applied to the drawer content container                 |
+| expanded   | N/A*                   | Styles applied to the root element when the drawer is expanded |
```
joebochill, line 107: Same comment as above...take it as a general comment on all of these NAs.

docs/Drawer.md
```diff
@@ -90,17 +90,33 @@ The `Drawer` has four `variant`s:
...
-| expanded   | Styles applied to the root element when the drawer is expanded |
-| paper      | MUI Drawer style override for the underlying paper element     |
-| sideBorder | Styles applied when sideBorder is set to true                  |
+```tsx
+import { Drawer } from '@brightlayer-ui/react-components';
+...
+<Drawer variant='temporary' sx={{margin: '4px'}}
+>
+</Drawer>
+```
```
joebochill, line 119: Why is this example here? The sections usually go Description, Usage/Example, API, Style Overrides (and there's already an example above).

docs/StyleOverridesGuide.md
```diff
@@ -0,0 +1,227 @@
...
+    icon={<CheckCircle />}
+    statusColor={Colors.green[500]}
+    iconColor={Colors.green[500]}
+/>;
+
+```
+
+## 3. Overriding Styles with the `className` and `classes` prop
+
+If you want to override a component's styles using custom classes, you can use the `className` prop, available on each component.
```
joebochill, line 73: This should say something about the `className` only being applied to the root element of the component.

docs/UserMenu.md
```diff
@@ -74,14 +74,21 @@ const avatar = <Avatar><SendIcon/></Avatar>;
...
+-   passing a `classes` prop with keys from the `Name` column below
+-   using the `Global CSS Class` in your main stylesheet
+
+For more details on styling options check out our [Styling Guide](https://github.com/brightlayer-ui/react-component-library/tree/master/docs#style-guide).
+
+| Name        | Global CSS Class         | Description                               |
+| ----------- | ------------------------ | ----------------------------------------- |
+| root        | .BluiUserMenu-root       | Styles applied to the root element        |
+|             | .BluiUserMenu-avatarRoot | Styles applied to the avatarRoot element  |
+| bottomSheet | n/a                      | Styles applied to responsive bottom sheet |
```
joebochill, line 91: This should still have a global class, but should have a callout saying that this element lives in a separate container and can't be styled via sx nesting rules.

components/src/core/Drawer/DrawerFooter.tsx
```diff
@@ -1,78 +0,0 @@
-import React, { HTMLAttributes } from 'react';
```
joebochill, line 1: Just to make sure that we haven't accidentally introduced a breaking change here related to the folder restructuring, did we check to make sure that all of these components that were relocated can still be individually imported from their original (prior) location?

e.g.:

```tsx
import { DrawerFooter } from '@brightlayer-ui/react-components/core/Drawer/DrawerFooter';
```

components/package.json
```diff
@@ -1,19 +1,20 @@
...
     "name": "@brightlayer-ui/react-components",
-    "version": "6.0.0",
+    "version": "6.1.0",
     "description": "React components for Brightlayer UI applications",
     "scripts": {
         "test": "jest src",
         "build": "tsc --p tsconfig.lib.json",
         "watch:storybook": "tsc --p tsconfig.lib.json --watch --outdir ../demos/storybook/node_modules/@brightlayer-ui/react-components",
-        "lint": "eslint \"src/**/**.{tsx,ts}\"",
+        "lint": "eslint \"src/**.{tsx,ts}\"",
```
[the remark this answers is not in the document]
joebochill: If you undo this change, there's 2 lint errors that need to be addressed.

components/package.json
```diff
@@ -1,19 +1,20 @@
...
         "lint:fix": "eslint \"src/**/**.{tsx,ts}\" --fix",
         "prettier": "prettier \"src/**/**.{ts,tsx,js,jsx,json,css,scss,html}\" --write",
         "prettier:check": "prettier \"src/**/**.{ts,tsx,js,jsx,json,css,scss,html}\" --check"
     },
     "prettier": "@brightlayer-ui/prettier-config",
     "dependencies": {
-        "@brightlayer-ui/colors": "^3.0.0",
+        "@brightlayer-ui/colors": "3.0.1",
+        "@emotion/css": "^11.9.0",
```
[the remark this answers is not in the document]
joebochill: Now that I'm doing a little more reading, why are we switching to CX over CLSX? 

Comment on an issue from one of the MUI developers:
image

Is our class merging logic still working as expected where we have replaced clsx with cx? It seems like it might be doing more under the hood than we were expecting.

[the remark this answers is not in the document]
joebochill: MUI doesn't use CX anywhere in their code (but they use CLSX a lot)

components/package.json
```diff
@@ -29,22 +30,24 @@
...
         "@types/enzyme": "^3.10.4",
         "@types/enzyme-adapter-react-16": "^1.0.5",
-        "@types/jest": "^27.4.1",
-        "@types/react-dom": "^17.0.11",
+        "@types/jest": "^27.0.0",
+        "@types/react-dom": "^18.0.5",
         "@typescript-eslint/eslint-plugin": "^5.14.0",
         "@typescript-eslint/parser": "^5.14.0",
         "@wojtekmaj/enzyme-adapter-react-17": "^0.6.0",
         "clsx": "^1.1.0",
```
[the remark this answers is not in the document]
joebochill: (it's a dependency of MUI, so we should pick it up for free)

components/src/core/Drawer/DrawerNavGroup/DrawerNavGroup.tsx
daileytj: Agreed that this is weird, most likely just following the pattern that's been there, but I'm happy to clean this up.

components/src/core/InfoListItem/InfoListItem.tsx
daileytj: Who's motive?! I don't even know who did this work at this stage in the game. 

gitblame suggests @bkarambe about 2 months ago ¯\_(ツ)_/¯ 

@bkarambe any thoughts on this?

components/package.json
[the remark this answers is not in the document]
emclaug2: Colors & Colors Branding just need to be demo'd to Joe before they can get published; otherwise they are in a good spot.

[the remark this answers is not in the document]
joebochill: Looks like that CX comment is within the context of using TSS.

I'm not opposed to using it, if it works...We probably should take 10 minutes and test out a few things to make sure the classes getting passed in via className, classes, and the default classes are getting merged together in the correct way (I think they are).

I'm also not opposed to continuing to use clsx since we don't have to add a new dependency and that's how MUI is continuing to do it today.

components/src/core/Drawer/DrawerNavGroup/DrawerNavGroup.tsx
joebochill: I _think_ I may have been the one who originally commented that spread out and I think it had something to do with spreading some of the props that are available in DrawerNavItem but not in DrawerRailItem. So my suggestion might not be valid...might need to test that specific case to see if the spread causes any issues.

components/src/core/InfoListItem/InfoListItem.tsx
joebochill: I mean, it's fine. I'm just curious.

bkarambe: > Who's motive?! I don't even know who did this work at this stage in the game.
> 
> gitblame suggests @bkarambe about 2 months ago ¯_(ツ)_/¯
> 
> @bkarambe any thoughts on this?

primaryTypographyProps was there previously, so just added the other styles. We can clean this up to pass through css.

components/src/core/Drawer/DrawerNavGroup/DrawerNavGroup.tsx
daileytj: After some testing, this works fine for the standard rail items but not those passed in via the items prop on `DrawerNavGroup`. I'm going to leave this as is for now.

components/src/core/InfoListItem/InfoListItem.tsx
daileytj: Leaving this as is since it's working right now. We can revisit this later if we think we can optimize this a bit by moving the style declaration to styles instead of using the typography props

components/src/core/Drawer/DrawerFooter.tsx
daileytj: this works out just fine!

components/src/core/AppBar/AppBar.tsx
```diff
@@ -1,50 +1,53 @@
 import React, { useCallback, useEffect, useState } from 'react';
-import { Theme, useTheme } from '@mui/material/styles';
-import createStyles from '@mui/styles/createStyles';
-import makeStyles from '@mui/styles/makeStyles';
+import { styled, useTheme } from '@mui/material/styles';
 import useMediaQuery from '@mui/material/useMediaQuery';
 import MuiAppBar, { AppBarProps as MuiAppBarProps } from '@mui/material/AppBar';
 // import { useScrollPosition } from '@n8tb1t/use-scroll-position';
 import clsx from 'clsx';
```
emclaug2, line 6: @daileytj I see you've removed the clsx dependency from the package.json, do we still have access to it through a transitive dependency? 

Considering that it builds in CI, I think this should be fine.

emclaug2 (approved):

ektasawant (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## v6.1.0 (June 22, 2022)
```
daileytj, line 3:
```suggestion
## v6.1.0 (June 24, 2022)
```

PR #461, opened 2022-07-13
Added react-dev folder to rcl and added script

bkarambe:
Fixes # BLUI- 3274

- Added react-dev folder and moved react dev docs to react component library
- Added a script to root package for starting the react dev project

#### To Test:
- yarn start:reactdev

- New folder is named as : **react-dev** and is kept inside **demos** folder. The script to run from root location is **yarn start:reactdev**
**Note: Preview page is just for testing the dummy api and playground to use redux**


emclaug2:
Change seems reasonable to me; 
If this new demo builds, could you add a step to our circleci config that runs that build step?

joebochill (commented):
Couple of comments:

- (organization) this isn't a demo project, it's our documentation site. So I don't think it makes sense to have this in the demos folder (makes more sense to me for it to replace the /docs folder — and then you don't have to duplicate all of the markdown files
- (organization) tabPanel and temporaryDrawer are not pages, so they probably don't belong in the /pages folder. They probably fit better in some sort of /navigation folder or in with the router setup.
- Why are we doing the ListItemTag stuff when the story was about Drawer (the idea was to start with one of the most complex components to identify issues and solve the hard problems early)?

joebochill:
Fresh clone also doesn't run to me (node-sass/python issue — this might be a local thing for me, but I haven't had issues running other projects):
image

emclaug2:
> I have moved react dev project to docs/ folder I have moved tabPanel and temporaryDrawer to router/navigation folder 

Can you make sure all your changes are pushed out?  TabPanel & TempDrawer are still inside the pages folder.

emclaug2 (approved):

PR #473, opened 2022-08-08
Added copy code button and functionality to dev docs example

bkarambe:
Fixes # BLUI-3289

- Added copy code functionality tto react dev docs Drawer Nav Item example

#### Screenshots:
Screenshot 2022-08-08 at 2 34 49 PM

#### To Test:
- yarn start:reactdev
- Navigate to Drawer->Drawer Nav Item
- Click on 'Copy All' button to copy the code snippet


jeffvg (approved):
looks good to me

docs/src/components/CopyToClipboardButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+    Omit<Omit<Omit<Omit<Omit<TooltipProps, 'placement'>, 'open'>, 'children'>, 'onMouseEnter'>, 'onMouseLeave'>,
+    'title'
+> & {
+    duration?: number;
+    position?: Position;
+    title?: string;
+    copiedPosition?: Position;
+    copiedTitle?: string;
+    copyText: string;
+};
```
huayunh, line 20: Hmm... what's the reason for expanding `TooltipProps` here? To follow MUI's tradition, "nested" types are usually defined like this:

```tsx
type CopyToClipboardProps = {
    title?: string;
    duration?: number;
    ... 
    TooltipProps?: TooltipProps;
};
```

And then you pass in any props related to the tooltip in an object, like this:

```tsx
<CopyToClipboard TooltipProps={{ content goes here }) ... />
```

docs/src/components/CopyToClipboardButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+export const CopyToClipboard: React.FC<CopyToClipboardProps> = (props) => {
+    const {
+        duration = 1000,
+        position = 'bottom',
+        title = '',
+        copiedPosition = 'bottom',
```
huayunh, line 26: We would never want to have a tooltip appear at two different positions in one component. `position` and `copiedPosition` is redundant.

docs/src/components/CopyToClipboardButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+                            sx={{
+                                fontSize: 16,
+                                cursor: 'pointer',
+                            }}
+                            onMouseEnter={(): void => {
+                                if (!isCopied) setShowTooltip(true);
+                            }}
+                            onMouseLeave={(): void => {
+                                if (!isCopied) setShowTooltip(false);
+                            }}
```
huayunh, line 55: you shouldn't be using `sx` and these event listeners in an icon component. These state controls should be handled in the parent component `<Button />`

docs/src/shared/utilities.ts
```diff
@@ -0,0 +1,25 @@
+function fallbackCopyTextToClipboard(text: string): void {
```
huayunh, line 1: If you are getting this code snippet from Stackoverflow, you are required to attribute the original author by at least linking back to the source answer, as all their answers are licensed under CC-BY-SA 4.0. We don't want to be plagiarizing.

huayunh (changes requested):

emclaug2:
@huayunh I haven't checked figma but are there specs for how you wanted the copy button to look when copied?  Angular is doing something completely different from this implementation.

docs/src/shared/utilities.ts
emclaug2: Not sure why this fallback implementation is necessary - we don't support Internet Explorer & the browser's `navigator.clipboard.writeText` method should work on all major browsers: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText

huayunh:
I see your point. I am actually flexible either way since neither of your designs poses any usability problems. I do not see a strong need to change and make two sides look identical.

docs/src/components/CopyToClipboardButton.tsx
bkarambe: I copied code directly from doc-it. Forgot to move event listeners to wrapper

docs/src/components/CopyToClipboardButton.tsx
bkarambe: I removed copiedPosition. We will be just keeping position prop, if user wants to change the position of tooltip

docs/src/components/CopyToClipboardButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    return isMobile ? null : (
+        <Box sx={{ display: 'flex', justifyContent: 'flex-end' }}>
+            <Tooltip {...toolTipProps} title={isCopied ? copiedTitle : title} placement={position} open={showTooltip}>
+                <Button
+                    variant="outlined"
+                    sx={{
+                        fontSize: 16,
+                        cursor: 'pointer',
+                    }}
```
huayunh, line 32: you shouldn't need this `sx` thing here; the button style should automatically get applied via blui themes

huayunh (commented):
Everything else looks good to me

huayunh (approved):

PR #471, opened 2022-08-02
Feature/3323 stylesheet for mdx

bkarambe:
Fixes # BLUI-3323

#### Changes proposed in this Pull Request:
- Added markdownMapping and markdown.css

#### Screenshots:
Screenshot 2022-08-02 at 4 30 28 PM

#### To Test:
- yarn start:reactdev
- Navigate to Drawer->Drawer->API Docs (tab)

PR #479, opened 2022-08-09
Feature/3318 drawer sub header api

ektasawant:
Fixes 3318 .

- api docs for drawer sub header
- 

#### To Test:
- http://localhost:3000/components/drawer/drawer-sub-header/api-docs

### DIFF
--- a/docs/src/components/navigation/componentPreviewTabs.tsx
+++ b/docs/src/components/navigation/componentPreviewTabs.tsx
@@ -13,7 +13,7 @@ interface TabPanelProps {
     value: number;
 }
 
-const hidePlaygroudTabs = ['drawer-layout'];
+const hidePlaygroudTabs = ['drawer-layout', 'drawer-sub-header'];
 const docsTabs = ['examples', 'api-docs', 'playground'];
 
 function TabPanel(props: TabPanelProps) {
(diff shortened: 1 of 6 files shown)


huayunh (approved):

huayunh:
@bkarambe , can you do something to make this gray area pure white? Looks like it was not addressed in #471 

image

jeffvg (approved):
looks good, agree with @huayunh about footer

docs/src/components/navigation/componentPreviewTabs.tsx
```diff
@@ -73,7 +73,7 @@ const tabStyles = {
 const tabPanelContentStyles = {
     display: 'flex',
     flexDirection: 'column',
-    height: '100%',
+    height: (theme: Theme) => `calc(100vh - ${theme.spacing(8)})`,
```
huayunh, line 76: Height shall stay 100%. 

The thing to change should be minHeight, or else you get strange things like this:
 
image

ektasawant: > Height shall stay 100%.
> 
> The thing to change should be minHeight, or else you get strange things like this:
> 
> image

Thanks for this suggestion :). Done the changes also updated other readme with material UI URL.

PR #488, opened 2022-08-24
Feature/3398 add props for drawer

ektasawant:
Fixes 3398 .

- Added remaining props for Drawer
- Added shared Props in Other tabs
- Added generate snippet code functionality

#### Screenshots:
Screenshot 2022-08-24 at 1 17 11 PM

#### To Test:
- http://localhost:3000/components/drawer/playground


emclaug2:
`noLayout` = true is causing some visual issues on the page. 
image

emclaug2 (commented):
We also don't have any code highlight on the angular side; I don't think it's required because it's going to make things a lot more difficult further along.

On the angular side, if a prop has been reset to its default value, the prop no longer appears in the code snippet.  This variance will cause the highlighted section to constantly be updated, so I'd just remove the highlighting all together.  It's too difficult to maintain & too timely to test.

docs/src/__types__/index.ts
```diff
@@ -1,7 +1,14 @@
...
+    max: number;
+    step: number;
+};
+
 export type PropsType = {
     propName: string;
     inputType: string;
-    inputValue: boolean | string | string[] | [];
+    inputValue: boolean | number | string | string[] | [];
+    inputSets?: SliderInputType;
```
emclaug2, line 11: I don't understand why this variable is called `inputSets` when it contains rangeData for a slider.  Seems weird to me.

emclaug2 (commented):
image

This prop doesn't seem to be updating on the code snippet or the playground demo.

emclaug2:
`disableActiveItemParentStyles` not updating on code snippet / demo.

ektasawant:
> For this story I'm going to ignore _where_ the SharedProps are located for now, I think Huayun is working on a story to discuss where the these knobs will be placed.
> 
> One thing I've immediately noticed is a severe lag while selecting a color on the color picker. If you cannot fix this as part of this story it deserves a separate issue. Lag Color [ Lag Color ](https://user-images.githubusercontent.com/6538289/186434771-4c7d288a-29ac-4413-9c56-0457d29c6cdc.gif) [ ](https://user-images.githubusercontent.com/6538289/186434771-4c7d288a-29ac-4413-9c56-0457d29c6cdc.gif)

We have stroy for color picker https://jira-prod.tcc.etn.com/browse/BLUI-3407

ektasawant:
> `noLayout` = true is causing some visual issues on the page. image

As per our last discussion I need to disable this prop. I will update this.

docs/src/__types__/index.ts
ektasawant: > We also don't have any code highlight on the angular side; I don't think it's required because it's going to make things a lot more difficult further along.
> 
> On the angular side, if a prop has been reset to its default value, the prop no longer appears in the code snippet. This variance will cause the highlighted section to constantly be updated, so I'd just remove the highlighting all together. It's too difficult to maintain & too timely to test.

Remove line highlighting.

ektasawant: Renamed inputSets  => rangeData

jeffvg:
couple items I noticed.

The slider labels are always displayed.
image

Child nav items can not be selected to see the "disable active item parent style work"
image
image



setting ripple true/false doesn't seem to get set in the drawer. and notice the vertical scroll bar while clicking.

https://user-images.githubusercontent.com/37150565/186490365-27620d57-3b5d-42b7-873b-06b937770a4b.mov

ektasawant:
@jeffvg , @emclaug2 please check updated PR with following changes

[X] - noLayout prop is disable now. It needs to be true for playground since it is overriding main drawer props. But I have updated helperText for this prop as `default value is false`.
[X] - removed line highlight 
[X] - updated key name inputSets => rangeData
[X] - removed background color from sharedProps, since it is not applicable to `Drawer`.
[X] - Fixed disableActiveItemParentStyles prop issue
[] - Styling and implementation of color picker will be fix in another story.
[] - Will update slider's styling in another story once we finalised new design https://www.figma.com/file/hOk7yEWOdZPbGTF6kze1sO/Dev-Doc?node-id=21%3A70094
[X] - Fixed Ripple prop.
[] - Not able to reproduce vertical scroll issue

emclaug2:
All the props seem to be working for me now.  

One thing that would be helpful is to omit props when their default value (in API docs) has been applied.

emclaug2 (approved):

ektasawant:
> One thing that would be helpful is to omit props when their default value (in API docs) has been applied.

This needs to be add more efforts to break code snippet. If it is good to have feature then it will be good to add as a feature request.

PR #492, opened 2022-08-30
Feature/3416 header playground

ektasawant:
Fixes 3416 .

- Common component for showing knobs and updating its value
- Common component for showing preview of component
- Updating states of all components in `docs/src/redux/componentsPropsState.tsx`
- Update props temporary drawer with latest design.

Please ignore following:
- Colorpicker styling
- Alignment of props in temporary drawer
- Code snippet styling.

#### To Test:
- http://localhost:3000/components/drawer/drawer-header/playground
- http://localhost:3000/components/drawer/playground


emclaug2:
I haven't looked at the code yet but this is looking pretty good.

One question - are we going to update the code snippet to omit default prop values?

image

ektasawant:
> are we going to update the code snippet to omit default prop values?

Not for this PR. I need to find some solution how I can break this string https://github.com/brightlayer-ui/react-component-library/pull/492/files#diff-7b995ded7fb400591b6e5465b784ba5580d7bb147e81bf49faba432d876c8440R31

emclaug2:
> Not for this PR. I need to find some solution how I can break this string https://github.com/brightlayer-ui/react-component-library/pull/492/files#diff-7b995ded7fb400591b6e5465b784ba5580d7bb147e81bf49faba432d876c8440R31

To address this on the angular side, I made a service that accepts a `Prop` and spits out the code required.
https://github.com/brightlayer-ui/angular-component-library/blob/release/dev-docs/src/docs/services/playground/playground.service.ts

If the `Prop` value is equal to the default value, it returns an empty string.

ektasawant:
> > Not for this PR. I need to find some solution how I can break this string https://github.com/brightlayer-ui/react-component-library/pull/492/files#diff-7b995ded7fb400591b6e5465b784ba5580d7bb147e81bf49faba432d876c8440R31
> 
> To address this on the angular side, I made a service that accepts a `Prop` and spits out the code required. https://github.com/brightlayer-ui/angular-component-library/blob/release/dev-docs/src/docs/services/playground/playground.service.ts
> 
> If the `Prop` value is equal to the default value, it returns an empty string.

Hi @emclaug2 , thanks for above solution. I have updated code to remove default props please review.

.firebaserc
```diff
@@ -1,5 +0,0 @@
-{
```
emclaug2, line 1: Can you re-add this file?

jeffvg:
noticed the divider knob is placed in the code snippet but is not in DOM for example that I can see

image

emclaug2:
I appreciate the refactor, Ekta - adding future playgrounds looks a lot more streamlined & easier to understand.

Regarding Jeff's comment about the divider - the divider is rendered but it isn't rendered correctly; maybe this is a bug because it's a Header rendered outside of a Drawer?

emclaug2:
The ColorPicker also was a lot more responsive, not sure if that was related to the refactor though.

ektasawant:
> The ColorPicker also was a lot more responsive, not sure if that was related to the refactor though.

Hi @jeffvg and @emclaug2 , I have fixed divider issue. Please check.

emclaug2 (approved):
All knobs work and looks fine to me.

emclaug2:
I know there's some styling issues with how elements are formatted but there's a story to tweak that already.

PR #494, opened 2022-09-02
Feature/3418 drawer footer playground

ektasawant:
Fixes 3418 .

#### Changes proposed in this Pull Request:
- Footer Playground
- Footer hideContentOnCollapse prop issue fixed [here](https://github.com/brightlayer-ui/react-component-library/pull/495)
- **Need to merge after [Drawer Subheader playground PR]**(https://github.com/brightlayer-ui/react-component-library/pull/494)

#### Screenshots:
Screenshot 2022-09-02 at 3 23 18 PM

#### To Test:
- http://localhost:3001/components/drawer-footer/playground

PR #493, opened 2022-08-30
Feature/3417 drawer subheader playground

ektasawant:
Fixes 3417 .

#### Changes proposed in this Pull Request:
- Subheader playground

#### Screenshots:
Screenshot 2022-08-30 at 9 48 50 PM

#### To Test:
- components/drawer-sub-header/playground

PR #499, opened 2022-09-05
add  playground for Hero

ektasawant:
Fixes 3405 .

- 
- 
- 

#### To Test:
- http://localhost:3001/components/hero/playground


huayunh:
- [ ] 1. What's the point of that `icon:ReactNode` if there's only one selection under that dropdown?
- [ ] 2. There are currently two different `icon` knobs. Can we think of a better way to group them? Or no Channel Value props at all? Anything on the Miro board Evan put together?
- [ ] 3. For form fields that takes `number | string`,  I think a free form text field does the work better.
- [ ] 4. Where is the bucket name for "required props"?
- [ ] 5. You would want to show some error message when the required `label` field is empty

ektasawant:
> * [X]  1. What's the point of that `icon:ReactNode` if there's only one selection under that dropdown? - **Added one more icon.**
> * [X]  2. There are currently two different `icon` knobs. Can we think of a better way to group them? Or no Channel Value props at all? Anything on the Miro board Evan put together? - **following https://brightlayer-ui-components.github.io/react/?path=/story/components-hero--with-full-config . Added valueIcon as new prop name (inside other props)**
> * [X]  3. For form fields that takes `number | string`,  I think a free form text field does the work better. - **Number field will work fine here. If we are using string need to concat `px` for every text.**
> * [ ]  4. Where is the bucket name for "required props"? - **Will fix this here BLUI 3471**. Looks like the design has change earlier there was no name label for required props. https://user-images.githubusercontent.com/10433274/187355246-5d238155-ab3c-4c6c-83ad-a427fafa1768.png
> * [ ]  5. You would want to show some error message when the required `label` field is empty - **Will fix this here BLUI 3471**
> * [X]  6. your `getIcon` function [is actually returning `<Fan fontSize={'inherit'} htmlColor={Colors.white[50]} />`](https://github.com/brightlayer-ui/react-component-library/pull/499/files#diff-d3a0d37e0b26bae8f12f57a24ac1e4becd3a990b8c0013f32c7393eb8d6835d9R57), but this is not currently addressed in the generated code snippet. - Fixed please review
> * [ ]  7. The code snippet is leaving a bottom margin to the bottom edge of the view port. - **This has been fixed in https://github.com/brightlayer-ui/react-component-library/pull/493 (please ignore here)**

ektasawant:
Making this as a draft since need to merge first https://github.com/brightlayer-ui/react-component-library/pull/493 and https://github.com/brightlayer-ui/react-component-library/pull/494

huayunh:
With regarding to the `ChannelValueProp`-related knobs in the "others" bucket, I think there are two questions that need to be answered:

1. Do we want to show these channel value props for a "hero" playground
2. if we do, what would be the best way to communicate the difference between knobs for channel value and knobs for the hero component itself

I have my own answers on these, but I would like to hear from you first

ektasawant:
> 2\.



> With regarding to the `ChannelValueProp`-related knobs in the "others" bucket, I think there are two questions that need to be answered:
> 
> 1. Do we want to show these channel value props for a "hero" playground
> 2. if we do, what would be the best way to communicate the difference between knobs for channel value and knobs for the hero component itself
> 
> I have my own answers on these, but I would like to hear from you first

 1. Do we want to show these channel value props for a "hero" playground - Hero has prop name `ChannelValueProps` so I think we need to let user know how to pass data to `ChannelValueProps `. We can hardcode this object with value / or use other Props section and inside helperText we can mention these are channel value props.
> 2. if we do, what would be the best way to communicate the difference between knobs for channel value and knobs for the hero component itself  - 
1. we can show it under Other props with proper helperText which will mention these props are from channel Value
2. Or use it under shared Props.
3. I do not want to show it under Channel Value Props section since this can be again need some changes in shared playground component :(

huayunh:
> “I think we need to let user know how to pass data to ChannelValueProps”

I found it fairly straightforward how data / props are passed to `ChannelValueProps`. I think you are correct in that we need to show the hero component is taking Channel Value's props, but I don't think Channel Value's prop is influencing significantly how Hero is behaving.

> "I do not want to show it under Channel Value Props section since this can be again need some changes in shared playground component :("

Two comments from me: 
* first, you should not let "how difficult it is to re-implement something" bias your design choice. I understand that this tends to be a major concern from the dev side, but I believe one should make the correct decision, instead of the easier one.
* second, I would argue that from an information hierarchy standpoint, putting these props under its own bucket/section is a better choice. Given the way these codes are structured, leaving it inside a helper text is not going to put enough emphasis. I regard helper text as a description / **supplement** to the text field, not a mean to **navigate** users.

ektasawant:
> > “I think we need to let user know how to pass data to ChannelValueProps”
> 
> I found it fairly straightforward how data / props are passed to `ChannelValueProps`. I think you are correct in that we need to show the hero component is taking Channel Value's props, but I don't think Channel Value's prop is influencing significantly how Hero is behaving.
> 
> > "I do not want to show it under Channel Value Props section since this can be again need some changes in shared playground component :("
> 
> Two comments from me:
> 
> * first, you should not let "how difficult it is to re-implement something" bias your design choice. I understand that this tends to be a major concern from the dev side, but I believe one should make the correct decision, instead of the easier one. 

- Not giving reason of Technical Limitations to bias design choice but if we are thinking to show section as per component. i should fix that in separate story. Otherwise since there are many playground stories are there in each story we may find something needs to update shared component. So my suggestion can be we can log an issue for new changes. And then we can finalize which playground needs changes. And then I can fix this.



> * second, I would argue that from an information hierarchy standpoint, putting these props under its own bucket/section is a better choice. Given the way these codes are structured, leaving it inside a helper text is not going to put enough emphasis. I regard helper text as a description / **supplement** to the text field, not a mean to **navigate** users.

huayunh:
I sure would honor that if that ended up being our design choice.

I have stated my opinion. Just to be sure, are you preferring to leave those Channel Value Props in their own section?

huayunh (approved):
Looks good to me.

docs/src/componentDocs/Hero/playground/HeroTypes.tsx
```diff
@@ -0,0 +1,86 @@
...
+                max: 100,
+                step: 1,
+            },
+        },
+        {
+            propName: 'label',
+            inputType: 'string',
+            inputValue: 'Velocity',
+            propType: 'string',
+            helperText: 'The text shown below the `ChannelValue`',
```
huayunh, line 44:
```suggestion
            helperText: 'The text shown below the Channel Value',
```

PR #502, opened 2022-09-06
Feature/3425 componentize copy code and GitHub button

bkarambe:
Fixes # BLUI-3425

- Added Copy code and github button inside Codeblock component
- Added Github button to Drawer nav item

#### Screenshots:
Screenshot 2022-09-07 at 12 35 30 AM

#### To Test:
- yarn start:reactdev


daileytj:
Be sure to think about if we have an example where we want to show the code block without either button, or with just one of the buttons. Or what if I want to use the button row separate from a code block. The way this is implemented doesn't allow for it and demands a url for the github link as well. If I was doing this I would have a component for the "code block" separate from the "button row" that conditionally renders each button if a url/copy text is provided. You can then conditionally render the entire component within the code block if either of those props are available. Or keep them separate and the examples would have to add the "code block" and the "button row".

bkarambe:
> @bkarambe Be sure to think about if we have an example where we want to show the code block without either button, or with just one of the buttons. Or what if I want to use the button row separate from a code block. The way this is implemented doesn't allow for it and demands a url for the github link as well. If I was doing this I would have a component for the "code block" separate from the "button row" that conditionally renders each button if a url/copy text is provided. You can then conditionally render the entire component within the code block if either of those props are available. Or keep them separate and the examples would have to add the "code block" and the "button row".

Done

emclaug2:
If you're looking for stuff to componentize, angular has made a component wrapper for all the live-examples on the examples page.

For example:

```
<Box sx={{ m: '32px 0', backgroundColor: colors.white[600], p: 4 }}>
    <DrawerNavItemWithIconsExample />
</Box>
```

That 32px margin & 4rem padding is duplicated on every example.  If this isn't a part of this story, I'd recommend making an issue for it instead.

docs/src/componentDocs/DrawerFooter/examples/ComplexDrawerFooter.tsx
```diff
@@ -39,9 +39,9 @@ export const ComplexDrawerFooter = (): JSX.Element => (
...
         <ComplexDrawerFooterExample />
         <CodeBlock code={codeSnippet} language="jsx" dataLine="9-30" />
-        <Box sx={{ display: 'flex', justifyContent: 'flex-end' }}>
-            <CopyToClipboard title={'Copy All'} copyText={codeSnippet} />
-            <FullCodeOnGithub sx={{ ml: 2 }} url="componentDocs/DrawerFooter/examples/ComplexDrawerFooterExample.tsx" />
-        </Box>
+        <CodeBlockActionButtonRow
+            copyText={codeSnippet}
+            url="componentDocs/DrawerFooter/examples/ComplexDrawerFooterExample.tsx"
+        />
```
daileytj, line 45: No need to do it now, but we may want to pull all these urls out into one config file and then read them here. This would allow us to modify them all in one place as a source of truth. Also makes for batch find and replace.

docs/src/shared/CodeBlockActionButtonRow.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import Box from '@mui/material/Box';
+import { CopyToClipboard } from './CopyToClipboardButton';
+import { FullCodeOnGithub } from './FullCodeOnGithubButton';
+import { SxProps } from '@mui/material/styles';
+
+export type CodeBlockActionButtonRowProps = {
+    shouldRenderCopyButton?: boolean;
+    shouldRenderGithubButton?: boolean;
```
daileytj, line 9: No need to add these props. You can just check if copytext or a url are present. If they are, render the respective button.

docs/src/shared/PreviewComponentWithCode.tsx
```diff
@@ -35,15 +34,6 @@ const PreviewComponentWithCode: React.FC<PreviewComponentProps> = (props): JSX.E
...
                 <CodeBlock code={code} language="jsx" />
-                <Box
-                    sx={{
-                        position: 'absolute',
-                        bottom: '16px',
-                        right: '400px',
-                    }}
-                >
-                    {show && <CopyToClipboard title={'Copy All'} copyText={code} />}
-                </Box>
```
daileytj, line 46: You could have left your button row in here and passed the copytext and url through here and only render the button row if one of those 2 props are available. That would save you a bunch of copy paste.

bkarambe:
> If you're looking for stuff to componentize, angular has made a component wrapper for all the live-examples on the examples page.
> 
> For example:
> 
> ```
> <Box sx={{ m: '32px 0', backgroundColor: colors.white[600], p: 4 }}>
>     <DrawerNavItemWithIconsExample />
> </Box>
> ```
> 
> That 32px margin & 4rem padding is duplicated on every example. If this isn't a part of this story, I'd recommend making an issue for it instead.

An issue has been logged https://github.com/brightlayer-ui/react-component-library/issues/508

docs/src/componentDocs/DrawerFooter/examples/ComplexDrawerFooter.tsx
bkarambe: [BLUI-3487](https://jira-prod.tcc.etn.com/browse/BLUI-3487)

PR #504, opened 2022-09-07
Dev Doc Drawer Nav Group Playground

ektasawant:
Fixes 3420 .

- Rename drawerTyps = drawerConfig same for other playground Json files.
- Update nav group playground
- 

#### To Test:
- http://localhost:3000/components/drawer/drawer-nav-group/playground

Note: for content projection prop like `titleContent` will add an example.


huayunh (approved):
Looks excellent! (Didn't check the code)

PR #526, opened 2022-09-15
Feature/3421 playground for rail item

ektasawant:
Fixes 3421 .

- Added Playground for Rail Item

**`activeItemFontColor`, `activeItemIconColor` not working in component library**. Issue Link  [527](https://github.com/brightlayer-ui/react-component-library/issues/527)

#### To Test:
- /components/drawer-rail-item/playground


ektasawant:
`activeItemFontColor`, `activeItemIconColor` not working in component library. Will raise an issue.

emclaug2:
I reran the deploy action so we can view it on firebase.

emclaug2:
This is weird - I've set the activeItemId to be 'Locations' yet the Title item is still active as well.
image

emclaug2:
This demo is confusing me a little. 

> This is weird - I've set the activeItemId to be 'Locations' yet the Title item is still active as well. image

To avoid confusion I would hardcode 'Title' as the active item in the demo and remove the knob.

emclaug2:
Can you also throw `backgroundColor` onto that new issue you've created?  This knob is doesn't affect the demo.

ektasawant:
> This demo is confusing me a little.
> 
> > This is weird - I've set the activeItemId to be 'Locations' yet the Title item is still active as well. image
> 
> To avoid confusion I would hardcode 'Title' as the active item in the demo and remove the knob.

Done removed ItemId from knob.

ektasawant:
> Can you also throw `backgroundColor` onto that new issue you've created? This knob is doesn't affect the demo.

Updated issue https://github.com/brightlayer-ui/react-component-library/issues/527

huayunh:
- [ ] 1. To avoid shocking our users, I recommend not using blue500 as the default background color on load... maybe white. Or "undefined", once we have that color picker properly implemented
- [ ] 2. Optional props — `condensed`. This is a prop better controlled by `<Drawer>` and I am having a mixed feeling placing it in a "Drawer Nav Rail Item". Here's how Joe reacted to it:
image

  - I think both solutions proposed by Joe works for me, but if you choose to apply it at the `<Drawer>` level, you need to make it clear in your knob description: "how condensed nav items without labels (will be applied to the parent Drawer component)"
  - If you choose to apply this to all Drawer Nav Items, make sure to take out the "(rail variant only)" from this knob's helper text — You are showcasing rail and only rail in this playground

- [ ] 3. Code Snippet — A standard drawer should have its default height set in `<Drawer>`, instead of having flex styles in `<DrawerBody>`. `<DrawerBody>` should not need the `backgroundColor={"transparent"}>` part either
- [ ] 4. It looks like this rail will never have an active rail item, leaving our effort showing `activeItemBackgroundColor`, `activeItemFontColor` and `activeItemIconColor` in vain. I recommend adding some event listener that sets the active item. 
- [ ] 5. `disableRailTooltip` prop, I would mention that it only get applied when the rail item is condensed
- [ ] 6. React's convention is to write `<DrawerRailItem ... divider={true} />` as `<DrawerRailItem ... divider />`

ektasawant:
> * [x]  1. To avoid shocking our users, I recommend not using blue500 as the default background color on load... maybe white. Or "undefined", once we have that color picker properly implemented - **added `transparent`**
> * [x]  2. Optional props — `condensed`. This is a prop better controlled by `<Drawer>` and I am having a mixed feeling placing it in a "Drawer Nav Rail Item". Here's how Joe reacted to it:
> 
> image
> 
> * I think both solutions proposed by Joe works for me, but if you choose to apply it at the `<Drawer>` level, you need to make it clear in your knob description: "how condensed nav items without labels (will be applied to the parent Drawer component)"
> * If you choose to apply this to all Drawer Nav Items, make sure to take out the "(rail variant only)" from this knob's helper text — You are showcasing rail and only rail in this playground - **Added condensed prop to every rail item**
> * [x]  3. Code Snippet — A standard drawer should have its default height set in `<Drawer>`, instead of having flex styles in `<DrawerBody>`. `<DrawerBody>` should not need the `backgroundColor={"transparent"}>` part either - **done**
> * [x]  4. It looks like this rail will never have an active rail item, leaving our effort showing `activeItemBackgroundColor`, `activeItemFontColor` and `activeItemIconColor` in vain. I recommend adding some event listener that sets the active item. - **now if we click any railItem it will be active. This `activeItemBackgroundColor` prop will work but for other 2 logged an issue https://github.com/brightlayer-ui/react-component-library/issues/527**
> * [x]  5. `disableRailTooltip` prop, I would mention that it only get applied when the rail item is condensed - **done**
> * [ ]  6. React's convention is to write `<DrawerRailItem ... divider={true} />` as `<DrawerRailItem ... divider />`: will fix this in BLUI-3543

docs/src/componentDocs/DrawerRailItem/playground/DrawerRailItemConfig.tsx
```diff
@@ -0,0 +1,141 @@
...
+            required: false,
+        },
+        {
+            propName: 'title',
+            inputType: 'string',
+            inputValue: 'Title',
+            propType: 'string',
+            helperText: 'The text to show on the first line',
+            required: true,
+            disable: true,
```
huayunh, line 68: why is this one also disabled?

docs/src/componentDocs/DrawerRailItem/playground/DrawerRailItemConfig.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+export const drawerRailItemConfig: ComponentType = {
+    componentName: 'Drawer Rail Item',
+    props: [
+        {
+            propName: 'condensed',
+            inputType: 'boolean',
+            inputValue: false,
+            propType: 'boolean',
+            helperText: 'Show condensed nav items without labels (rail variant only)',
```
huayunh, line 12:
```suggestion
            helperText: 'Show condensed nav items without labels',
```

docs/src/componentDocs/DrawerRailItem/playground/DrawerRailItemConfig.tsx
```diff
@@ -0,0 +1,141 @@
...
+            propType: 'string',
+            helperText: 'Status stripe and icon color',
+            required: false,
+        },
+        {
+            propName: 'title',
+            inputType: 'string',
+            inputValue: 'Title',
+            propType: 'string',
+            helperText: 'The text to show on the first line',
```
huayunh, line 66:
```suggestion
            helperText: 'The label text',
```

huayunh: I realized that we are calling it `label`, `title` and `item text`

huayunh: I will make a separate issue/PR to update these in the API Doc

docs/src/componentDocs/DrawerRailItem/playground/DrawerRailItemConfig.tsx
```diff
@@ -0,0 +1,141 @@
...
+            helperText: 'Whether to show a line between all items',
+            required: false,
+            defaultValue: false,
+        },
+        {
+            propName: 'itemFontColor',
+            inputType: 'colorPicker',
+            inputValue: Colors.black[500],
+            propType: 'string',
+            helperText: 'The color used for the item text',
```
huayunh, line 118:
```suggestion
            helperText: 'The color used for the title text',
```

docs/src/componentDocs/DrawerRailItem/playground/DrawerRailItemConfig.tsx
```diff
@@ -0,0 +1,141 @@
...
+            required: false,
+            defaultValue: false,
+        },
+        {
+            propName: 'icon',
+            inputType: 'select',
+            inputValue: '<Device />',
+            options: ['<Device />', '<Devices />'],
+            propType: 'JSX.Element',
+            helperText: 'A component to render for the left icon',
```
huayunh, line 40:
```suggestion
            helperText: 'An icon component to render',
```

huayunh (commented):
everything else looks good. I like how the props are now applied to all drawer items

emclaug2:
Couple inquiries
 
1. Even after adding the `condensed` prop to every nav item, the width is still 72px instead of the expected 56.  Is there another issue to address this?

2. I like how the example is interactive on the playground, selecting different active items.  I see you've included a disabled input for activeItem ID, but it is now out-of-sync with the example.   What is the intension behind this?

huayunh:
> Even after adding the condensed prop to every nav item, the width is still 72px instead of the expected 56. Is there another issue to address this?

that's #541, an issue with the component library itself

docs/src/componentDocs/DrawerRailItem/playground/DrawerRailItemConfig.tsx
ektasawant: The reason to disable the title string is if we are changing the title then I need to update the `activeItem` of the drawer onChange of Title input. And this needs to be done in SharedDrawer which is difficult. We can do this but I guess the current playground changes working fine to show railItem props. Any thoughts?

ektasawant:
> Couple inquiries
> 
> 1. Even after adding the `condensed` prop to every nav item, the width is still 72px instead of the expected 56.  Is there another issue to address this?
> 2. I like how the example is interactive on the playground, selecting different active items.  I see you've included a disabled input for activeItem ID, but it is now out-of-sync with the example.   What is the intension behind this? - Removed ItemID from props.

jeffvg (approved):
looks good, will the shared color picker and condensed prop be cleaned up for MVP?

docs/src/componentDocs/DrawerRailItem/playground/DrawerRailItemConfig.tsx
huayunh: Doesn't `activeItem` correspond to only `itemID`, not `title`?

ektasawant:
> looks good, will the shared color picker and condensed prop be cleaned up for MVP?

For condensed prop, we have story https://jira-prod.tcc.etn.com/browse/BLUI-3531 in TODO. And shared color picker will be part of BLUI-3552 (if Joe add this point in story description)

huayunh:
@ektasawant see my reply above https://github.com/brightlayer-ui/react-component-library/pull/526#discussion_r980058568

docs/src/componentDocs/DrawerRailItem/playground/DrawerRailItemConfig.tsx
ektasawant: I missed this comment, didn't get a notification of this comment. To fix the activeItem issue I have made changes in this PR https://github.com/brightlayer-ui/react-component-library/pull/553. If it looks fine I can fix a rail item similar to this PR.

PR #530, opened 2022-09-15
Feature/channel value playground

ektasawant:
Fixes 3502 .

- Update channelValue playground
- Fix copy all button from playground snippet

#### To Test:
- http://localhost:3000/components/channel-value/playground


huayunh (changes requested):
`unitSpace` is actually not required. Would you update both API Table and Playground to reflect this?

Things look good otherwise.

ektasawant:
> `unitSpace` is actually not required. Would you update both API Table and Playground to reflect this?
> 
> Things look good otherwise.

We have this prop inside channelValue component. https://github.com/brightlayer-ui/react-component-library/blob/dev/components/src/core/ChannelValue/ChannelValue.tsx#L57. Any specific reason to remove it from API docs and playground?

huayunh:
Not to remove it, but to update the API table so that it does not say unitSpace is required

ektasawant:
> Not to remove it, but to update the API table so that it does not say unitSpace is required

Oops now I understand . Let me update.

ektasawant:
> https://blui-react-docs--pr530-feature-channel-valu-om8r7o61.web.app

Screenshot 2022-09-16 at 1 53 10 PM
Screenshot 2022-09-16 at 1 53 23 PM

Done.

ektasawant:
> `unitSpace` is actually not required. Would you update both API Table and Playground to reflect this?
> 
> Things look good otherwise.

Merged this story.

huayunh (commented):
lgtm

(didn't check code)

PR #572, opened 2022-10-01
Implement custom color picker and number picker components

joebochill:
~This has an associated JIRA story, but I don't have the number for it since I'm not on the network at the moment.~
Fixes BLUI-3407

- Implement a ColorPicker component
    - accepts free text or picking color from the dropdown 
    - supports css colors ('red'), hex colors ('#fff', '#ffff', '#ffffff'), rgb/rgba, hsl/hsla
    - supports mui palette strings (configurable, disabled by default) ('primary.main', 'grey.600', 'text.disabled', 'common.main', 'action.hover', etc.)
    -  Validates color input and shows error message
    - Debounces onChange event so it doesn't get spammed with redux calls
- Implement a NumberPicker component
    - accepts keyboard input and has increment/decrement buttons
    - clicking the input will open the slider panel
- Add style rule to use Roboto for the input values
- Fix the linking script to copy components to docs project
- Fix issue where InfoListItem throws/crashes when it receives an invalid background color
- Ran Lint and Prettier

##### Color Picker
https://user-images.githubusercontent.com/29152776/193419488-f6f22df4-7a3a-4212-b9d6-48829867b732.mov

##### Number Picker
https://user-images.githubusercontent.com/29152776/193427460-c17ff15b-d9a4-47d5-80f3-bf885269d730.mov

#### To Test:
- Launch Dev Docs
- Go to the ChannelValue Playground
- Go nuts on the color picker & number picker


huayunh:
I haven't checked the code yet, but the demo looks wonderful

ektasawant:
Working fine for most of the components. For InfolistItem only for the `BackgroundColor` input,  gives the below error.
Screenshot 2022-10-03 at 5 08 18 PM

joebochill:
We'll need to make the 'mui-style colors' capability toggle-able. Some of our components/props will not work with a mui-style color value (they will only work if the style is applied through `sx`). A number of our components apply styles through 'styled-components' which do not use the same syntax. We should update all of our components so they can use these shorthand color values (https://mui.com/system/styled/#how-can-i-use-the-sx-syntax-with-the-styled-utility) but that isn't realistic with our current timeframe, so simply disabling that capability on the components/props that it doesn't work for will be fine.

huayunh:
> We'll need to make the 'mui-style colors' capability toggle-able

do you plan to address it in this PR?

huayunh:
# 1

Add a border to that color stripe. A MUST.

image

# 2 
When the color value is not recognized, the color picker icon changes to a question mark

image

# 3 
A few knobs are listed to accept `number|string` but are treated like a pure `number` type. Maybe you have your own thoughts on what design decisions to make here.
image

# 4
A few knobs, such as `titleColor` in Drawer Nav Group and a lot of Drawer shared props are better left blank to show their "default" state and help accelerate dev doc in dark theme implementation. Feel free to log an issue.

# 5

When a knob is set to undefined, its code should be toggled off
image

joebochill:
I did 1 and 2.

For number 3, I'm inclined to leave that for now. If you want to be able to do both in one, we probably need a different design for that (showing a slider for a string is weird). For this, I was just building a number input. 
Additionally, just because a prop will accept multiple types of input doesn't necessarily mean we have to let people input all of those different types in the playground (numeric is the more typical use case).

For 4 and 5, I won't touch those in this PR. Those are outside the scope of the input component and are more related to the logic that takes the value from the input and puts it into the right places in the redux store and the code snippet.

huayunh:
Can you log issues for post MVP clean-up for 4 and 5?

joebochill:
4 and 5 are mentioned in my cleanup audit notes

huayunh (approved):

PR #583, opened 2022-10-06
Feature/3552 remove style overrides name quotes

bkarambe:
Fixes # BLUI-3552

API Docs
In the 'style overrides' section, the Name should not be in a code block (second bullet)
Do we need to update the 'any other props will be provided to the root component' to Box instead of div? I'm pretty sure in the MUI 5 we converted everything to use box.
All tables should be 100% width

#### Screenshots:
Screenshot 2022-10-06 at 2 41 53 PM

#### To Test:
- yarn strat:reactdev


emclaug2:
@bkarambe Can you please resolve these merge conflicts?

emclaug2:
Also see your other PR for the CI github deploy fix.

The summary is that Firebase pushed out a buggy package & this change manually sets the firebase cli version instead of using latest.

huayunh:
why does the drawer layout API doc table look so different from other API tables?

huayunh:
Somehow, when I navigate to Toolbar Menu Examples page, there is a console warning. I don't know if this is caused by this PR or not.

```
Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>.
```

bkarambe:
> Somehow, when I navigate to Toolbar Menu Examples page, there is a console warning. I don't know if this is caused by this PR or not.
> 
> ```
> Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>.
> ```

Logged an issue https://github.com/brightlayer-ui/react-component-library/issues/586

bkarambe:
> why does the drawer layout API doc table look so different from other API tables?

Good catch @huayunh . There were some styles applied to table long back. Removed now.

Issue #515, opened 2022-09-12
App Bar animationDuration prop not expanding app bar when set to 1000ms


ektasawant:
#### Describe the bug
App Bar `animationDuration` prop not expanding the app bar when set to 1000ms.
#### What is the expected behavior?
App Bar `animationDuration` prop should expand the app bar when set to 1000ms. 
#### What are the steps to reproduce?
1. Go to dev docs `/components/app-bar/playground`
2. set animationDuration to 1000ms
3. scroll window app bar is not expanding.(This is inconsistent )

#### Environment

All browsers	
#### Anything else to add?
https://github.com/brightlayer-ui/react-component-library/pull/511#issuecomment-1241330853
https://github.com/brightlayer-ui/react-component-library/pull/511#issuecomment-1243547676

Issue is related to Component Library.

joebochill:
The 1000ms part of this issue is irrelevant now. Closing this and creating a new issue to track the inconsistent collapse/expand behavior described above: https://github.com/etn-ccis/blui-react-component-library/issues/736

PR #604, opened 2022-10-13
Fix/3552 playground issues

ektasawant:
Fixes 3552 .

**Common for all playground**
- Be consistent with 'undefined' in the dropdowns…it should either be always first (my preference) or always last…right now different inputs do it differently
- Don't include undefined props
- Nice to have: don’t include props if they are set to the default value

**Appbar Playground** 

- Default variant to snap
- When the appbar collapses, the code window is resizing and the example grows — these should be fixed sizes so the collapsing looks normal
- AnimationDuration doesn't seem like it's working…or the range of values is not wide enough
- scrollThreshold needs a wider range of values
- Don’'t show scrollContainerId knob

**Channel value Playground**

- Add another icon option (TrendUp/TrendDown)

**Drawer Playground**

- activeItem should be a dropdown with the ids that are used in the example
- Don't show the noLayout prop
- Temporary drawer is not contained to the demo window and can never be closed
- Rail/condensed are broken

**Drawer Header Playground**

- Don't show active item - it blocks the view of the divider

**Drawer Sub Header Playground**

- Don't have an activeItem selected
- 'Others' props should not be styled as a prop if they are not a prop of the component (Open)…this should probably just say 'Drawer Open'

**Drawer Footer Playground**

- 'Others' props should not be styled as a prop if they are not a prop of the component (Open)…this should probably just say 'Drawer Open'

**User Menu Playground**

- UseBottomSheetAt prop is not functional

- 
- 

#### To Test:
-


daileytj:
These issues still remain:

### AppBar Playground
- AnimationDuration doesn't seem like it's working…or the range of values is not wide enough

### Drawer Playground
- Temporary drawer is not contained to the demo window and can never be closed

ektasawant:
> These issues still remain:
> 
> ### AppBar Playground
> * AnimationDuration doesn't seem like it's working…or the range of values is not wide enough - https://github.com/brightlayer-ui/react-component-library/issues/515 there is an issue log in the component library that need to fix in another PR.
> 
> ### Drawer Playground
> * Temporary drawer is not contained to the demo window and can never be closed
For this, I have a story https://jira-prod.tcc.etn.com/browse/BLUI-3545

Hi @daileytj, sorry I forgot to remove the above 2 points.

daileytj (approved):
@ektasawant Approving this assuming the things I commented on will be addressed in a near-future PR

ektasawant:
> @ektasawant Approving this assuming the things I commented on will be addressed in a near-future PR

Yes, for 1st I need to investigate but for 2nd, we have a story in this sprint.

Issue #591, opened 2022-10-11
App Bar not taking default value if collapsedHeight or expandedHeight is empty


ektasawant:
#### Describe the bug
App Bar not taking default value if collapsedHeight or expandedHeight is empty.
#### What is the expected behavior?
App Bar should tak default value if collapsedHeight or expandedHeight is empty.
#### What are the steps to reproduce?
1. Go to Dev docs - /components/app-bar/playground
2. remove number from  collapsedHeight prop i.e pass empty string
3. remove number from  expandedHeight prop i.e pass empty string

The app bar starts misbehaving (Check below recording)

#### Screenshots / Screen recording
https://user-images.githubusercontent.com/8997218/194061258-0762d624-a980-4e8e-9c56-e43cda9406ad.mov
#### Code snippet / Link to minimum reproduction example

```
[CODE HERE](https://github.com/brightlayer-ui/react-component-library/blob/dev/components/src/core/AppBar/AppBar.tsx#L85)
https://github.com/brightlayer-ui/react-component-library/blob/dev/components/src/core/AppBar/AppBar.tsx#L92
```

#### Your environment information

All

daileytj:
This issue has been fixed

PR #606, opened 2022-10-17
fix rail item and nav item playground

ektasawant:
Fixes 3552 .

**Drawer Nav Item Playground**

- Don't show itemID
- Have more than one icon option in the dropdown
- This playground is confusing because some of the knobs apply to all of the nav items and some of them only apply to the first one - **As discussed removed duplicate props from child nav item since child nav item inherits shared props of parent nav**

**Drawer Rail Item Playground**

- Condensed should be set at the Drawer level
- Icon input is non-functioning
- Make title prop editable

#### To Test:
- http://localhost:3000/components/drawer/drawer-nav-item/playground

- http://localhost:3000/components/drawer-rail-item/playground


huayunh:
Drawer Nav Item Playground:
- [ ] By default the icon should not be `undefined`
- [ ] Divider knob's helper text says "Whether to show a line between **all** items". The divider should only be applied to the root items.

ektasawant:
> Drawer Rail Item Playground:
> 
> * [x]  The activeBackgroundColor, can you either make the default empty, or set it to blue500 at 5%? For a second I thought the drawer nav rail was broken
> * [x]  Any chance to put Drawer Props above the Shared Props?
> * [ ]  The helper text of "statusColor" says "Status stripe and icon color". There is no icon involved here. - This description is in our component library seems `statusColor` prop apply to icon and statusStripe both.

https://github.com/brightlayer-ui/react-component-library/blob/dev/components/src/core/InfoListItem/InfoListItem.tsx#L95
https://github.com/brightlayer-ui/react-component-library/blob/dev/components/src/core/InfoListItem/InfoListItem.tsx#L169

ektasawant:
> Drawer Nav Item Playground:
> 
> * [x]  By default the icon should not be `undefined`
> * [ ]  Divider knob's helper text says "Whether to show a line between **all** items". The divider should only be applied to the root items. - Divider know applied to all nav items except nested one - We can use helperText like this "Whether to show a line between all items except nested items"

Screenshot 2022-10-19 at 6 11 09 PM

huayunh:
> This description is in our component library seems `statusColor` prop apply to icon and statusStripe both.

This is the Drawer Rail Item component. It does not use Info List Item. 

https://github.com/brightlayer-ui/react-component-library/blob/dev/components/src/core/Drawer/DrawerRailItem/DrawerRailItem.tsx#L299

It looks like our API doc is misaligned on this too. Would you mind patching that as well?

ektasawant:
> > Divider know applied to all nav items except nested one - We can use helperText like this "Whether to show a line between all items except nested items"
> 
> Maybe "whether to show a line between non-nested items"? Trying to avoid saying the word "items" twice in a short sentence.

Done.

huayunh:
Looks good.

I noticed that when I cleared the style inputs, the drawer active nav item is not showing the default styles for active items. Haven't these already been fixed? The drawer rail item looks good.

image

huayunh:
I also noticed that the drawer is currently placed inside a `<Box>` instead of `<DrawerLayout>` in the generated code snippet.

ektasawant:
> Looks good.
> 
> I noticed that when I cleared the style inputs, the drawer active nav item is not showing the default styles for active items. Haven't these already been fixed? The drawer rail item looks good.
> 
> image

This is an issue similar to [591](https://github.com/brightlayer-ui/react-component-library/issues/591) empty string not handled by the component. Will fix this in the playground for now. And log an issue in component library.

ektasawant:
> > Looks good.
> > I noticed that when I cleared the style inputs, the drawer active nav item is not showing the default styles for active items. Haven't these already been fixed? The drawer rail item looks good.
> > image
> 
> This is an issue similar to [591](https://github.com/brightlayer-ui/react-component-library/issues/591) empty string not handled by the component. Will fix this in the playground for now. And log an issue in component library.

I tried to fix it from Dev docs but had no success. Logged an issue here https://github.com/brightlayer-ui/react-component-library/issues/611

huayunh (approved):

PR #609, opened 2022-10-19
match drawer playground with angular

ektasawant:
Fixes 3545 .

- Update drawer playground with angular [layout playground](https://brightlayer-ui-components.github.io/angular-dev/components/drawer/layout/playground)
- 
- 

#### To Test:
- /components/drawer/playground

### DIFF
--- a/docs/src/componentDocs/Drawer/playground/PreviewComponent.tsx
+++ b/docs/src/componentDocs/Drawer/playground/PreviewComponent.tsx
@@ -93,135 +93,133 @@ export const PreviewComponent = (): JSX.Element => {
         variantIsRail
             ? `<Box>App Content Here.</Box>`
             : `<Box>
-        <AppBar position="static">
-             <Toolbar>
-                 <IconButton
-                     size="large"
-                     edge="start"
-                     color="inherit"
-                     aria-label="menu"
-                     sx={{ mr: 2 }}
-                     onClick={(): void => ${toggleOpenProp(drawerProps.open)}}
-                 >
-                     <Menu />
-                 </IconButton>
-                 <Typography variant="h6">Toolbar</Typography>
-             </Toolbar>
-         </AppBar>
-         <Box>App Content Here.</Box>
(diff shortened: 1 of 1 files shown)


huayunh:
this PR suffers from the same issue described here https://github.com/brightlayer-ui/react-component-library/pull/606#issuecomment-1284294290

huayunh:
I also noticed that the drawer is currently placed inside a `<Box>` instead of `<DrawerLayout>` in the generated code snippet.

ektasawant:
> I also noticed that the drawer is currently placed inside a `<Box>` instead of `<DrawerLayout>` in the generated code snippet.

Since this is not a drawerLayout playground not showing in the snippet. The same was discussed with @joebochill.

joebochill:
If the `Box` isn't doing anything, I would not show that in the snippet either, just the relevant code for the Drawer.

ektasawant:
> If the `Box` isn't doing anything, I would not show that in the snippet either, just the relevant code for the Drawer.

Done.

huayunh:
"just the relevant code for the Drawer" — I suppose that means no page body either, only what's inside `<Drawer>`

docs/src/componentDocs/Drawer/playground/PreviewComponent.tsx
```diff
@@ -185,95 +218,133 @@ export const PreviewComponent = (): JSX.Element => {
...
+                                    </Toolbar>
+                                </AppBar>
+                                <Box
+                                    sx={{
+                                        p: 2,
+                                        ml: drawerProps.open
+                                            ? `${drawerProps.width}px`
+                                            : drawerProps.variant === 'persistent'
+                                            ? 7
+                                            : 1,
```
huayunh, line 339: Temporary drawers should be "content-cover" instead of "content-push" — that is, when a temporary drawer is open, it should hover above the content, not push to the side.

huayunh (approved):
Change looks good otherwise

PR #625, opened 2022-11-01
Fix overflow on playground

joebochill:
- Fix overflow on playground caused by updated Tabs position value
  - the strange numbers are because the tab bar renders at 49px tall
- Fix issue where tabs overflow under the nav drawer on medium size screens

#### Screenshots:
image

#### To Test:
- Verify that the playground does not scroll vertically
- Verify that the tabs are all fully visible across all screen sizes

### DIFF
--- a/docs/src/shared/PreviewComponentWithCode.tsx
+++ b/docs/src/shared/PreviewComponentWithCode.tsx
@@ -17,11 +17,10 @@ const PreviewComponentWithCode: React.FC<PreviewComponentProps> = (props): JSX.E
             <Box
                 sx={{
                     p: 2,
-                    height: '70vh',
+                    height: { xs: 'calc(70vh - 105px)', sm: 'calc(70vh - 113px)' },
                     display: 'flex',
                     justifyContent: 'center',
                     alignItems: 'center',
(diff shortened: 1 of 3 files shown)


docs/src/shared/PreviewComponentWithCode.tsx
```diff
@@ -17,11 +17,10 @@ const PreviewComponentWithCode: React.FC<PreviewComponentProps> = (props): JSX.E
             <Box
                 sx={{
                     p: 2,
-                    height: '70vh',
+                    height: { xs: 'calc(70vh - 105px)', sm: 'calc(70vh - 113px)' },
```
daileytj, line 20: Do you want to add a comment as to what these specific calculations are doing? Where does the 105 & 113 px come from?

daileytj (approved):
Approving, assuming we're cool with those height calculations.

jeffvg (approved):

docs/src/shared/PreviewComponentWithCode.tsx
joebochill: Done

PR #645, opened 2022-11-23
Feature/blui 3630 rtl tests

jeffvg:
Fixes BLUI-3630.

- remove enzyme and enzyme-adapters
- add react test library and testing-library/jest-dom
- update tests to use render, screen and cleanup
- update attributes for data-test={'foo'} to data-testid={'blui-foo'} so RTL & jestDOM find test ids
- use codecov orb in CI config (codecov@3.2.2)
- remove codecov deprecated dependency
- set CODECOV_TOKEN in CircleCI project

#### To Test:
- yarn test


components/src/core/AppBar/AppBar.tsx
```diff
@@ -287,6 +287,7 @@ const AppBarRender: React.ForwardRefRenderFunction<unknown, AppBarProps> = (prop
         <Root
             ref={ref}
             {...muiAppBarProps}
+            data-testid={'appbar-root'}
```
emclaug2, line 290: I'm thinking these `data-testid` tags need to be prefixed with `blui-` in the event that there's a duplicated test id in our adopters' apps.

components/src/core/ChannelValue/ChannelValue.test.tsx
```diff
@@ -1,49 +1,56 @@
...
             </ThemeProvider>
         );
+        expect(screen.getByText('test')).toBeTruthy();
     });
     it('should render with the wrapper class', () => {
-        const wrapper = mountWithTheme(<ChannelValue value={1} />, theme);
-        expect(findByTestId('wrapper', wrapper)).toBeTruthy();
-        wrapper.unmount();
+        render(
+            <ThemeProvider theme={theme}>
```
emclaug2, line 24: What is the difference between this test vs the one above it, "renders without crashing"?

components/src/core/Drawer/Drawer.test.tsx
```diff
@@ -21,144 +19,162 @@ import * as BLUIThemes from '@brightlayer-ui/react-themes';
...
+        //     'b_1_0',
+        //     'c',
+        //     'c_0',
+        //     'c_0_0',
+        //     'c_0_1',
+        // ];
+
+        // const navItemList = wrapper.find(InfoListItem);
+        // expect(navItemList.length).toEqual(expectedNavItemTitleList.length);
+        // navItemList.forEach((item, index) => {
```
emclaug2, line 170: Test should be commented out if no assertions, or assertions should be uncommented out.

components/src/core/EmptyState/EmptyState.tsx
```diff
@@ -83,7 +83,7 @@ const EmptyStateRender: React.ForwardRefRenderFunction<unknown, EmptyStateProps>
         <Root
             ref={ref}
             className={cx(defaultClasses.root, classes.root, userClassName)}
-            data-test={'frame'}
+            data-testid={'frame'}
```
emclaug2, line 86: Why "frame" vs something like "blui-empty-state-root"?  Also I'm a little confused about these test ids; our internal tests won't do this, but app teams can have multiple elements rendered on screen at the same time which would then share an id.   I'm wondering if this will cause any complications down the road.  Will it be our recommendation to adopters that they can also query elements using our internal test ids?  

Can these testing ids be queried as if they are classes?  I know cypress uses [data-cy] which doesn't assume the id is unique.

components/src/core/AppBar/AppBar.tsx
jeffvg: by default react test library looks for data-testid and changing them to be blui-data-testid or data-blui-testid breaks the findByTestId methods. I'm sure there are config options for this but would be a hack

components/src/core/EmptyState/EmptyState.tsx
jeffvg: left this as is and didn't change "frame" to something more meaningful. Having duplicate data-testid needs research but having them would require the test to focus on the first data-testid using "firstChild" or second....open to suggestions.

components/src/core/ChannelValue/ChannelValue.test.tsx
jeffvg: removed dup test

emclaug2 (approved):

huayunh:
Sorry I've been out of the loop with UI testing. Why are we commenting out those tests in drawer?

jeffvg:
> Sorry I've been out of the loop with UI testing. Why are we commenting out those tests in drawer?

because they need refactored to work with RTL, I'm looking for help with these

huayunh:
I wrote some of the tests you commented out. I can try, but I doubt I will be very useful in this PR

jeffvg:
> I wrote some of the tests you commented out. I will give it a try, but I doubt I will be very useful in this PR

ok, we will have new stories to update these and to add additional later. so we can wait until then or go for it

huayunh:
sounds good with me

huayunh:
Alright @jeffvg I got 3 out of the 4 commented-out tests working. I don't know how to make that "inherits and overrides properties from Drawer" one to detect child props.

jeffvg:
> Alright @jeffvg I got 3 out of the 4 commented-out tests working. I don't know how to make that "inherits and overrides properties from Drawer" one to detect child props.

good deal! thank you

PR #654, opened 2022-12-07
Publish v6.1.2

daileytj:
- Publishes v6.1.2


emclaug2 (approved):

jeffvg:
all looks good but do we need to do anything in the style overrides MD file here?

https://github.com/brightlayer-ui/react-component-library/blob/master/docs/src/shared/markdown/StyleOverridesGuide.md#3-continue-using-jss-until-it-is-fully-removed-from-muimaterial

daileytj:
@jeffvg I discussed this with @joebochill and we decided to leave it in. The instructions are still valid if they use a certain mui/material version, but it's also far down on our list of recommendations so hopefully no one is using it.

jeffvg (approved):

PR #666, opened 2022-12-27
Bugfix/3651 364 right component and chevron together

### LINKED ISSUE #364: Add Ability to Use chevron and rightComponent prop on the InfoListItem at the same time
#### Describe the bug
On the InfoListItem, if you specify `rightComponent` it will override the chevron and only show the right component.

User should be able to show the chevron and a custom rightComponent without having to manually combine them in the rightComponent prop.

#### What are the steps to reproduce?
1. clone the repository
2. `yarn start:showcase`
3. Go to the list item example: http://localhost:3000/brightlayer-ui-components/data-display-components
4. Configure an ILI to have `chevron={true}` and `rightComponent={/*something*/}`
5. Observe that you only see the rightComponent
		
#### Anything else to add?
We have a pseudo workaround already in place within the DrawerNavItem. It is doing the work of combining the chevron and a rightComponent into one item for rightComponent. We should be able to base the approach around this, but in the ILI instead.

When fixing this, we will also need to do some refactoring in the Drawer components so that chevron and rightComponent are able to be passed down independently to the ILI.

bkarambe:
Fixes #364 BLUI-3651

- Rendered chevron and right component separately in info list item
- Refactored drawer nav item to pass chevron, chevronColor and rightComponent props 

#### Screenshots:
Screenshot 2022-12-27 at 1 21 46 PM

#### To Test:
- Checkout [bugfix/i3651-364-infolistitem-chevron-right-comp](https://github.com/brightlayer-ui/react-showcase-demo/tree/bugfix/i3651-364-infolistitem-chevron-right-comp) branch in react showcase demo
- build the components with yarn build
- paste the dist folder contents to showcase demo's node_modules/@brighlayer-ui/react-components/
- cd demos/showcase
- yarn start

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,28 +4,24 @@
 
 ### Fixed
 
--   DOM nesting console warnings in `<InfoListItem>`'s `title`, `subtitle` and `info` props ([#644](https://github.com/brightlayer-ui/react-component-library/issues/644)).
--   Rendering of chevron and right component in `<InfoListItem>` at the same time ([#364](https://github.com/brightlayer-ui/react-component-library/issues/364)).
-
-### Updated
-
--   `<DrawerNavItem>` to pass `chevron` and `chevronColor` props to `<InfoListItem>` ([#364](https://github.com/brightlayer-ui/react-component-library/issues/364)).
+-   DOM nesting console warnings in `<InfoListItem>`'s `title`, `subtitle` and `info` props ([#644](https://github.com/etn-ccis/blui-react-component-library/issues/644)).
+-   Rendering of chevron and right component in `<InfoListItem>` at the same time ([#364](https://github.com/etn-ccis/blui-react-component-library/issues/364)).
 
 ## v6.1.2 (December 7, 2022)
 
 ### Updated
 
--   Remove dependency on @mui/styles ([#633](https://github.com/brightlayer-ui/react-component-library/issues/633)).
+-   Remove dependency on @mui/styles ([#633](https://github.com/etn-ccis/blui-react-component-library/issues/633)).
 
 ## v6.1.1 (November 1, 2022)
 
 ### Fixed
 
--   `hideContentOnCollapse` prop of `<DrawerFooter>` not hiding footer content ([#484](https://github.com/brightlayer-ui/react-component-library/issues/484)).
--   Temporary drawer rendering, due to incorrectly passing open prop ([#486](https://github.com/brightlayer-ui/react-component-library/issues/486)).
--   `activeItemFontColor`, `activeItemIconColor` prop of `<DrawerRailItem>` not updating `font color` and `icon color` of active rail item. ([#486](https://github.com/brightlayer-ui/react-component-library/issues/486)).
--   `condensed` prop of `<DrawerRailItem>` not appling height and width `56X56` to all rail items. ([#541](https://github.com/brightlayer-ui/react-component-library/issues/541)).
--   Styling of `nonClickableIcon` in `<DrawerHeader>` not applying properly ([#562](https://github.com/brightlayer-ui/react-component-library/issues/562)).
+-   `hideContentOnCollapse` prop of `<DrawerFooter>` not hiding footer content ([#484](https://github.com/etn-ccis/blui-react-component-library/issues/484)).
+-   Temporary drawer rendering, due to incorrectly passing open prop ([#486](https://github.com/etn-ccis/blui-react-component-library/issues/486)).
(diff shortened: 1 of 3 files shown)


components/src/core/InfoListItem/InfoListItem.tsx
```diff
@@ -193,20 +193,24 @@ const InfoListItemRender: React.ForwardRefRenderFunction<unknown, InfoListItemPr
...
+                    <InfoListItemChevron
+                        chevronColor={chevronColor}
+                        color={'inherit'}
+                        role={'button'}
+                        className={combine('chevron')}
+                    />
+                )}
+            </>
+        ),
+        [rightComponent, chevron, combine]
```
huayunh, line 212: I was a bit surprised that the `rightComponent` prop is not necessary for this dependency list while dev doc playground still works.

huayunh (commented):
Can verify everything works as intended.

It looks like this is an issue in the Angular side also and _maybe_ in React Native as well. Is there a story / bug to address that? If not, would you please log an issue / bug?

huayunh:
I have mixed feelings about the drawer side though. I am leaning towards keeping the original behavior — when a drawer nav item's `chevron` is checked, it should not affect the behavior of the `expandIcon` / `collapseIcon`. Thoughts @joebochill ?

Original | This PR
---|---
image|image

bkarambe:
> Can verify everything works as intended.
> 
> It looks like this is an issue in the Angular side also and _maybe_ in React Native as well. Is there a story / bug to address that? If not, would you please log an issue / bug?

**Angular**
I checked on Angular side, we can add both right component and chevron at the same time.
Code:
Screenshot 2022-12-28 at 3 09 04 PM
Screen:
Screenshot 2022-12-28 at 3 06 33 PM

**React Native**
On react native, chevron will not be displayed if there is right component (similar to react).
Code:
Screenshot 2022-12-28 at 3 24 57 PM
Screen:
Screenshot 2022-12-28 at 3 25 32 PM

Logged an issue for react native https://github.com/brightlayer-ui/react-native-component-library/issues/312

components/src/core/InfoListItem/InfoListItem.tsx
joebochill: I'm not sure I understand this comment...rightComponent _is_ in the dependency array.

joebochill:
> I have mixed feelings about the drawer side though. I am leaning towards keeping the original behavior — when a drawer nav item's `chevron` is checked, it should not affect the behavior of the `expandIcon` / `collapseIcon`. Thoughts @joebochill ?

I'm conflicted on this too. It would make sense that if the chevron prop is true, then the chevron should be there. But I would lean towards leaving the original behavior as well. Before if the DrawerNavItem had children or items, then the chevron prop is ignored and the expand/collapse icon is the only thing you see.

components/src/core/InfoListItem/InfoListItem.tsx
huayunh: my bad, meant to type `chevronColor`

joebochill: yeah, chevronColor should be in there. If it's working anyway, we're probably just getting lucky and the entire component is re-rendering based on some other prop change or something like that.

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 ### Fixed
 
 -   DOM nesting console warnings in `<InfoListItem>`'s `title`, `subtitle` and `info` props ([#644](https://github.com/brightlayer-ui/react-component-library/issues/644)).
+-   Rendering of chevron and right component in `<InfoListItem>` at the same time ([#364](https://github.com/brightlayer-ui/react-component-library/issues/364)).
+
+### Updated
+
+-   `<DrawerNavItem>` to pass `chevron` and `chevronColor` props to `<InfoListItem>` ([#364](https://github.com/brightlayer-ui/react-component-library/issues/364)).
```
huayunh, line 12: I recommend taking this log item out. You do not need to log every single change you made, especially when they do not directly influence how our end users use it.

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 ### Fixed
 
 -   DOM nesting console warnings in `<InfoListItem>`'s `title`, `subtitle` and `info` props ([#644](https://github.com/brightlayer-ui/react-component-library/issues/644)).
+-   Rendering of chevron and right component in `<InfoListItem>` at the same time ([#364](https://github.com/brightlayer-ui/react-component-library/issues/364)).
```
huayunh, line 8: I recommend updating **all** issue links in this changelog to point to their new location in the new etn-ccis org. It should be a quick find-and-replace.

```diff
- https://github.com/brightlayer-ui/react-component-library/issues/364
+ https://github.com/etn-ccis/blui-react-component-library/issues/364
```

huayunh (approved):
New change works

daileytj (approved):

PR #727, opened 2023-03-16
Added external links to doc it resources, roadmap, release notes to d…

surajeaton:
Fixes # BLUI-3305

-Added doc it links to roadmap, release notes and resources

Screenshot 2023-03-16 at 1 25 47 PM

#### To Test:
- yarn start:reactdev

-


huayunh:
I don't know if this will be too much to ask for this PR. The new tag is intended to be shown every time there is a new release. As soon as the user clicked on that button, the new tag will be gone until the next release is out. Similar to how docit handles the announcement banner above the app bar.

joebochill:
@JeffGreiner-eaton looks like we have an issue with the labeler

joebochill:
> I don't know if this will be too much to ask for this PR. The new tag is intended to be shown every time there is a new release. As soon as the user clicked on that button, the new tag will be gone until the next release is out. Similar to how docit handles the announcement banner above the app bar.

Let's skip this for now. It was not explicitly called out in the acceptance criteria and it's a large amount of effort compared to how the story was estimated based on expectations. Let's make a new story for it to avoid scope creeping this.

daileytj:
These buttons should be round...

It's easier to see when you click, but tough to grab a screenshot of my clicking without video sooo the hover screenshot will have to suffice for now
Screen Shot 2023-03-16 at 1 14 46 PM

JeffGreiner-eaton:
> @JeffGreiner-eaton looks like we have an issue with the labeler

Yes I see. Multiple commits in same PR causing PR labeler to get removed. Not sure what to do about it.

joebochill:
> > @JeffGreiner-eaton looks like we have an issue with the labeler
> 
> Yes I see. Multiple commits in same PR causing PR labeler to get removed. Not sure what to do about it.

Please create a JIRA story to investigate this. This is a very common use case that we need to work properly.

surajeaton:
@daileytj Please review.

huayunh (changes requested):
- [ ] You should not be using an icon button in this case. Do not force your users to click on a small button toward the end of your drawer item to navigate. The entire drawer item should be clickable; the external link thing should function only as a decoration, denoting that you are now taken to somewhere outside dev doc.
- [ ] You should not use `Color.gray[200]` in your code directly, as this (1) takes an extra step to import and (2) does not work in dark theme. Find the closest theme color and use it.
- [ ] Why is `titleDivider` set to `false`? Is that some specific design decision you made?
- [ ] Why are you modifying `drawer.tsx` instead of having your drawer navigation defined using `pageDefinitions` instead?

huayunh:
- [ ] Why are these components in a folder on their own? If there is only one file other than index.tsx, it is not necessary to create a folder there.
image

surajeaton:
> * [x]  Why are these components in a folder on their own? If there is only one file other than index.tsx, it is not necessary to create a folder there.
> 
> image

Removed ExternalLinkIcon component as we are not using it now.

surajeaton:
> * [x]  You should not be using an icon button in this case. Do not force your users to click on a small button toward the end of your drawer item to navigate. The entire drawer item should be clickable; the external link thing should function only as a visual indicator, denoting that the users are now being taken to somewhere outside dev doc.
> * [x]  You should not use `Color.gray[200]` in your code directly, as this (1) takes an extra step to import and (2) does not work in dark theme. Find the closest theme color and use it.
> * [ ]  Why is `titleDivider` set to `false`? Is that some specific design decision you made?
> * [ ]  Why are you modifying `drawer.tsx` instead of having your drawer navigation defined using `pageDefinitions` instead?

In designs, I did not find the title divider. That's why I kept it false. Removed 
Screenshot 2023-03-21 at 12 01 28 PM

As discussed in the last week's bug scrub, it is ideal to add drawer nav group and nav item to drawer.tsx, we need to add icon (currently right comp) to the title. For that we need to alter the typings. But it is okay to directly add those in the drawer.tsx

docs/src/router/drawer.tsx
```diff
@@ -136,6 +148,41 @@ export const NavigationDrawer: React.FC = () => {
                         items={convertNavItems(navGroup.pages || [], navGroup.path || '', 0, handleNavigate, dispatch)}
                     />
                 ))}
+                <DrawerNavGroup
+                    title="COMMUNITY"
+                    titleColor={Colors.blue[500]}
```
huayunh, line 153: is it possible to put `"primary.main"` here instead of a direct color reference?

huayunh (commented):
Behavior looks good

docs/src/router/drawer.tsx
surajeaton: We are not using sx prop in DrawerNavGroup instead we are using style, so passing a string will not parse. 
Screenshot 2023-03-23 at 1 47 29 PM

huayunh (approved):
very good!

PR #730, opened 2023-03-21
Use Roboto Mono and Stacks

huayunh:
- Use `'Roboto Mono'` whenever possible
- Use `<Stack>` whenever possible

### DIFF
--- a/docs/src/index.css
+++ b/docs/src/index.css
@@ -26,7 +26,7 @@ code[class*='language-'],
 pre[class*='language-'],
 code[class*='language-'] *,
 pre[class*='language-'] * {
-    font-family: 'Roboto Mono' !important;
+    font-family: 'Roboto Mono', monospace !important;
 }
 
 pre[class*='language-'] {
--- a/docs/src/shared/CodeBlock.tsx
+++ b/docs/src/shared/CodeBlock.tsx
@@ -33,7 +33,7 @@ export const CodeBlock: React.FC<CodeBlockProps> = (props): JSX.Element => {
                 data-line={dataLine}
                 style={{ margin: 0, width: '100%', backgroundColor: Colors.black[800], borderRadius: 4 }}
             >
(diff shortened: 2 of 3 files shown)


daileytj:
What's up with all the re-formatting? Was it not prettier before?

huayunh:
> What's up with all the re-formatting? Was it not prettier before?

I don't know. CI didn't pass the check, so I ran `yarn prettier` locally.

> What issue number or bug are you working from here? What prompted the `<Stack>` refactor?

None. `<Stack>` is a cleaner (and perhaps the more encouraged way by MUI) way to handle flex box than grouping everything in a `<Box>`

docs/src/index.css
```diff
@@ -26,7 +26,7 @@ code[class*='language-'],
 pre[class*='language-'],
 code[class*='language-'] *,
 pre[class*='language-'] * {
-    font-family: monospace !important;
+    font-family: 'Roboto Mono' !important;
```
daileytj, line 29: Should we add a monospace fallback here?

docs/src/shared/CodeBlock.tsx
```diff
@@ -33,7 +33,7 @@ export const CodeBlock: React.FC<CodeBlockProps> = (props): JSX.Element => {
                 data-line={dataLine}
                 style={{ margin: 0, width: '100%', backgroundColor: Colors.black[800], borderRadius: 4 }}
             >
-                <code style={{ fontFamily: 'monospace' }} className={`language-${language}`}>
+                <code style={{ fontFamily: 'Roboto Mono' }} className={`language-${language}`}>
```
daileytj, line 36: monospace Fallback?

daileytj (approved):

docs/src/shared/CodeBlock.tsx
daileytj:
```suggestion
                <code style={{ fontFamily: '"Roboto Mono", monospace' }} className={`language-${language}`}>
```

PR #747, opened 2023-04-03
Add a landing page to dev doc

huayunh:
Fixes I did this for fun

## Changes proposed in this Pull Request:

- Added a landing page
  - page can be visited by going to the root `/` url
  - can also be visited by clicking on the drawer header

## Screenshots / Screen Recording (if applicable)

https://user-images.githubusercontent.com/8997218/229401538-274e45a1-e46c-4d0a-a4ec-71463d5c63dc.mov


docs/src/pages/HomePage.tsx
```diff
@@ -0,0 +1,135 @@
...
+    pageRoot: {
+        background: (theme: Theme): string =>
+            `url('${CirclesImage}') no-repeat center/200%, ${theme.palette.primary.main}`,
+        height: '100%',
+        color: 'primary.contrastText',
+    },
+    content: {
+        maxWidth: CONTENT_WIDTH,
+        margin: (theme: Theme): string => `${theme.spacing(9)} auto`,
+        padding: `${PADDING}px`,
```
surajeaton, line 29: Instead of **padding** you can just use **p**.

docs/src/pages/HomePage.tsx
```diff
@@ -0,0 +1,135 @@
...
+const styles = {
+    pageRoot: {
+        background: (theme: Theme): string =>
+            `url('${CirclesImage}') no-repeat center/200%, ${theme.palette.primary.main}`,
+        height: '100%',
+        color: 'primary.contrastText',
+    },
+    content: {
+        maxWidth: CONTENT_WIDTH,
+        margin: (theme: Theme): string => `${theme.spacing(9)} auto`,
```
surajeaton, line 28: Instead of margin and passing theme, you can simply use:
my: 9
mx: 'auto'

docs/src/pages/HomePage.tsx
```diff
@@ -0,0 +1,135 @@
...
+    }, []);
+
+    return (
+        <Box sx={styles.pageRoot}>
+            <SharedAppBar title={''} />
+            <Stack sx={styles.content} gap={8}>
+                <Box sx={styles.pageTitle}>
+                    <Typography variant={lgUp ? 'h2' : 'h3'}>
+                        {lgUp && (
+                            <span
```
surajeaton, line 72: Instead of using span directly, you can use MUI's Box component and pass component prop as "span" and use "sx" instead of style:
 ```
<Box
      component={'span'}
      sx={{ position: 'relative', right: '1em', top: 'calc( 1em + 10px )', opacity: 0.5 }}
  >
      $
  </Box>
```

surajeaton (commented):
Not passing title to SharedAppBar looks odd:
Screenshot 2023-04-03 at 3 02 35 PM

docs/src/pages/HomePage.tsx
huayunh: The PADDING constant is set to 24, but `p` only takes the "unit size", like 3 instead of 24

docs/src/pages/HomePage.tsx
huayunh: Thanks. I don't see any direct value in using Box and forcing it to be a span instead of just using a span, but I can go either way.

PR #749, opened 2023-04-03
Bug/fix undefined image

JeffGreiner-eaton:
Fixes https://github.com/etn-ccis/blui-react-component-library/issues/717.

- Update getImage prop to be optional and set undefined
-
-

-

#### To Test:

-

-


huayunh (commented):
Feels like a hack when `url(undefined)` does not return 404. You did indeed get rid of the warning, but I'd prefer to see if you can get the `background-image` CSS rule to go away when `undefined` is selected by the user in the dropdown.

image

JeffGreiner-eaton:
> Feels like a hack when `url(undefined)` does not return 404. You did indeed get rid of the warning, but I'd prefer to see if you can get the `background-image` CSS rule to go away when `undefined` is selected by the user in the dropdown.
> 
> image

For now I thinks its ok as is and since the utilities > getImage call will be refactored in the upcoming sprints.. @daileytj  what do you think

daileytj:
> > Feels like a hack when `url(undefined)` does not return 404. You did indeed get rid of the warning, but I'd prefer to see if you can get the `background-image` CSS rule to go away when `undefined` is selected by the user in the dropdown.
> > image
> 
> For now I thinks its ok as is and since the utilities > getImage call will be refactored in the upcoming sprints.. @daileytj what do you think

I agree that this will be refactored, but it may be a while before we get to this so we should probably just fix it right now. Ideally, the getImage method call should not even get called unless there is an image.

huayunh (approved):

huayunh:
Nice!

PR #756, opened 2023-04-06
Use `process.env` for relative links

huayunh:

This link in the deployed environment doesn't work because it is a relative link without accounting for the `PUBLIC_URL`. I am injecting the correct base URL here:
image


JeffGreiner-eaton (approved):
looks good

PR #757, opened 2023-04-06
update gtag to use react app gaid

JeffGreiner-eaton:
With the current changes in the PR, I see no page views coming in running of the firebase deployment but I do see "users" increment on count and would need to actually test this config under a dev deployment to see what's going on. If anyone has ideas on this let me know

- update to use %REACT_APP_GAID% in index
-
-

-

#### To Test:

-

-

### DIFF
--- a/docs/public/index.html
+++ b/docs/public/index.html
@@ -57,7 +57,7 @@
         window.dataLayer = window.dataLayer || [];
         function gtag() { dataLayer.push(arguments); }
         gtag('js', new Date());
-        gtag('config', '%gaID%', { send_page_view: false })
+        gtag('config', '%REACT_APP_GAID%', { send_page_view: false })
     </script>
 </head>
 
--- a/docs/public/index.html
+++ b/docs/public/index.html
@@ -51,7 +51,7 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>Brightlayer UI React Docs</title>
-    <script async src="https://www.googletagmanager.com/gtag/js?id=%gaID%">
+    <script async src="https://www.googletagmanager.com/gtag/js?id=%REACT_APP_GAID%">
     </script>
     <script>
         window.dataLayer = window.dataLayer || [];


huayunh (commented):
I think this will yield two `<script>` tags in the `<head>` — one initiated by `public/index.html` and another by `src/index.tsx` and may cause issues since we are loading the same script twice. You may not need these two Google analytics `<script>` in `public/index.html`

joebochill:
> I think this will yield two `<script>` tags in the `<head>` — one initiated by `public/index.html` (the one you are changing) and another by `src/index.tsx` (the one currently in effect in the dev branch) — and may cause issues since we are loading the same script twice. You may not need these two Google analytics `<script>` in `public/index.html`

I was wondering this as well, if we truly need _both_ the script tags in the index.html _and_ the initialize call in index.tsx. If we were getting data when the scripts were not properly set, it makes me suspect there is probably some redundancy.

JeffGreiner-eaton:
> I think this will yield two `<script>` tags in the `<head>` — one initiated by `public/index.html` (the one you are changing) and another by `src/index.tsx` (the one currently in effect in the dev branch) — and may cause issues since we are loading the same script twice. You may not need these two Google analytics `<script>` in `public/index.html`

Yeah not sure what to think about this. Having 2 scripts shouldn't harm anything. We have 2 seperate items.
- First one is the wrapper

image

- Second is gTag
- 
image

joebochill:
As long as you followed the instructions for setting it up and that's what they said to do, i'm fine with it...it just seems curious. It shouldn't hurt anything unless somehow it starts double-counting things...but that should be pretty obvious when we look at the numbers.

JeffGreiner-eaton:
> As long as you followed the instructions for setting it up and that's what they said to do, i'm fine with it...it just seems curious. It shouldn't hurt anything unless somehow it starts double-counting things...but that should be pretty obvious when we look at the numbers.

So the setup assistant in GA4 initializes the new property from the old property, the very next step is to insert the google tag manager script and this is what has been done. I tried running only on ReactGA and ID like we have always done but it didn't seem to fetch as much data like it did before but once I used both google tag manager & ReactGA the data seemed more relevant to what we normally would see.  In all I would like to test this out as is in this PR just to see what we get and if we do not like the results we can revert back to using our original ReactGA and google wrapper

huayunh (approved):

PR #810, opened 2023-10-04
Update dev doc master branch

huayunh:

- Added language direction switcher
- Added a section to refer to the React workflow package
- Dependabot stuff

#### To Test:

- Verify that the language direction switcher is working
- Verify that the workflow page exists and is pointing to the correct page.


ektaghag-eaton:
RTL view breaks icons. Are we supposed to check only LTR to RTL is breaking or do we need to test UI as well?
For example Icon space is breaking and the drawer RTL for the multilayer menu is breaking.
Screenshot 2023-10-05 at 1 31 46 PM

joebochill:
As far as I can tell, this code is not ready for a master merge. The LTR switcher still has pending work items before we want to ship it (it's really janky as Ekta has pointed out). It also looks like there's a bunch of half-baked code around Playground component refactoring which is not used/not hooked up (at least it appeared that way during some of the pair programming sessions I sat in with Andrew). If you want to get your workflows update in, I would suggest doing it as a hotfix or isolating the other items (in a clearly marked folder, or moving them to an experimental branch) until they are finished.

huayunh:
👍 I think half of these are beyond our control and the other half require some form of tweaking within our component library.

huayunh:
Ok, I am not sure how to do a hotfix in this case... leave that for others

PR #817, opened 2023-11-06
Convert Component playground(s) to use playground library

joebochill:

- Convert following components' playground to use the library components from https://github.com/etn-ccis/blui-react-documentation-components
    - [x] ~~AppBar~~
    - [x] ~~ChannelValue~~
    - [x] Drawer
    - [x] DrawerHeader
    - [x] DrawerSubheader
    - [x] DrawerNavGroup
    - [x] DrawerNavItem
    - [x] DrawerRailItem
    - [x] DrawerFooter
    - [x] ~~EmptyState~~
    - [x] ~~Hero~~
    - [x] ~~InfoListItem~~
    - [x] ~~ListItemTag~~
    - [x] ScoreCard
    - [x] ThreeLiner
    - [x] ToolbarMenu
    - [x] UserMenu
- Remove unused files and redux state

image

#### To Test:

- `yarn start:docs`
- go to updated playground pages and play around with the props
- See if anything is broken (other than missing copy button)

- What did we miss in the shared library?

> Known Issues (fixes in progress): 
> - Support for object-style props is a little weak (e.g., ChannelValueProps in Hero)
> - Large component snippets do not scroll


joebochill:
This is a low-priority item — don't feel compelled to review this right now unless you've already been through everything else that's part of the current sprint.

docs/src/shared/ComponentPreviewTabs.tsx
```diff
@@ -146,7 +146,12 @@ export const ComponentPreviewTabs = (): JSX.Element => {
                 </Box>
             </TabPanel>
             <TabPanel value={value} index={2}>
-                <Box sx={playgroundContentStyles}>
+                <Box
+                    sx={[
+                        playgroundContentStyles,
+                        location.pathname.includes('channel-value') ? { marginRight: 0 } : {},
+                    ]}
+                >
```
huayunh, line 154: add a `// TODO` comment to remind yourself/others to clean up these when everything has been migrated.

docs/src/componentDocs/ChannelValue/playground/PlaygroundPage.tsx
```diff
@@ -1,17 +1,158 @@
...
+};
+
+const generateSnippet: CodeSnippetFunction = (data) =>
+    `<ChannelValue 
+    ${getPropsToString(getPropsMapping(data, inputConfig), { join: '\n\t', skip: ['icon', 'htmlColor'] })}
+    ${
+        data.icon && data.icon !== 'undefined'
+            ? `icon={${(data.icon as string).replace('/>', '')}fontSize={'inherit'}${
+                  data.htmlColor && data.htmlColor !== 'undefined' ? ` htmlColor={'${data.htmlColor as string}'}` : ''
+              } />}`
```
huayunh, line 144: `undefined` is a fairly common thing in our react components, given that we often want to test the component behavior when an optional prop is not given. Is it possible to incorporate this into the package? Like, for all `required: false` prop data, add an extra field `undefinedOption: true` or something like that?

huayunh (commented):
wow

docs/src/componentDocs/ChannelValue/playground/PlaygroundPage.tsx
joebochill: I think what you are describing is achievable already. Each parameter that you define to show in the sidebar has the following core values configurable:
image

So you can set the defaultValue that a prop has (for required or optional props).

The `'undefined'` string being used here is because this parameter renders as a select/dropdown and there is an explicit option in there for 'undefined' (without the option, there is no way to unselect an item once you've selected one). The select input type requires option `values` to be `number | string` — this could _probably_ be extended to support a value of `undefined` as well if we want to be able to do a comparison like `data.icon !== undefined` instead. That might complicate the typings a little bit (probably why I didn't do it initially)

joebochill: Now I think I remember — allowing `undefined` as a value in the dropdown would throw an error:
> Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component

joebochill: I had considered providing the ability to do a value mapping from string (label in dropdown) to anything (actual value you want to parameter to have, e.g. `undefined` or `<Component />`), like `options: [{label: 'undefined', value: undefined}]`, but that seemed to overcomplicate things for minimal gain. But we could revisit that if we want to. It might help when defining the preview component so the user doesn't have to do the mapping/lookup themselves.

huayunh: I think it can help with both 
* our testing (passing in the supposed default value vs. actually passing in undefined) and
* simplify this `generateSnippet` function that our adopters will write, like what you said there

joebochill: I took another look at it, and I think we'll leave it as-is for now. With the mapping capability, the user would still be stuck — e.g., In the case of the ChannelValue example, when picking an icon, the actual icon element they would pass as the value in the mapping would need to also know the value of the htmlColor input to render properly. Since the config object is static, they wouldn't be able to access that variable, so I don't think there's any real gain. That may not have made a lot of sense, but I can show you what I mean if you want to dig deeper.

huayunh (commented):
The style of this code. It should not start a new line
image

Suspicious amount of spacing on the bottom
image

Dark theme mobile, incorrect bottom drawer color when collapsed (looks fine when expanded)
image

Dark theme accordion has weird background colors:
image

joebochill:
> Suspicious amount of spacing on the bottom

This was intentional to allow users to edit field 'in their workspace' rather than trying to edit fields at the very bottom of the sidebar. This is a common feature of code/text editors where users are able to 'overscroll' past the last line of text to bring it up to the middle/top of the editor.

> Dark theme accordion has weird background colors
These are unstyled accordions (i.e., default style from the theme) — what color do you want them to be, same as the background?

huayunh:
> what color do you want them to be, same as the background?

the `paper` color should be fine.

joebochill:
Everything has been migrated to the new components now...this is ready for more scrutiny

huayunh (commented):
This `*` feels weird (Drawer / Drawer playground)
image

Dark theme, need to make the accordion and the background the same color (I don't care you set which to what, as long as the background color is the same)
image

Drawer header playground, the generated code snippet yields a mix of tabs and spaces. 

A few generated code snippet shows the closing `>` on a new line when no props are present, making my `prettier` head twitch. This is drawer subheader's playground, with its default configuration:
image

Empty state, selecting `<Router>` for the `icon` prop will cause the icon to disappear (it exists as a MUI Icon and a Material Design icon)
image

joebochill:
> This `*` feels weird (Drawer / Drawer playground)
> 
> Dark theme, need to make the accordion and the background the same color (I don't care you set which to what, as long as the background color is the same)
> 
> Drawer header playground, the generated code snippet yields a mix of tabs and spaces.
> 
> A few generated code snippet shows the closing `>` on a new line when no props are present, making my `prettier` head twitch. This is drawer subheader's playground, with its default configuration: 
> 
> Empty state, selecting `<Router>` for the `icon` prop will cause the icon to disappear (it exists as a MUI Icon and a Material Design icon) 

Can we log these as bugs so we don't have to keep pushing changes into a 130+ file PR?

huayunh (approved):

huayunh:
All issues have been logged. 

* https://github.com/etn-ccis/blui-react-component-library/issues/831
* https://github.com/etn-ccis/blui-react-component-library/issues/830
* https://github.com/etn-ccis/blui-react-component-library/issues/829
* https://github.com/etn-ccis/blui-react-documentation-components/issues/1
* https://github.com/etn-ccis/blui-react-documentation-components/issues/2

PR #850, opened 2024-02-12
update width calc

JeffGreiner-eaton:
Fixes https://github.com/etn-ccis/blui-react-component-library/issues/806.

- update width calc & max width
-
-

-

#### To Test:

- https://blui-react-docs--pr850-bug-blui-5256-mobile-s6pe8xfp.web.app/components/drawer/playground
- resize view using dev tools

- The drawer is using noLayout prop and default size is 360px and on small devices the item nav group chevron is cut off from view. Should the example be updated for this?
- Also wondering if a different calculation should be done here for Box component?

### DIFF
--- a/docs/src/componentDocs/Drawer/playground/PlaygroundPage.tsx
+++ b/docs/src/componentDocs/Drawer/playground/PlaygroundPage.tsx
@@ -246,7 +246,8 @@ const DrawerPreview: PreviewComponent = ({ data }) => {
                     sx={{
                         backgroundColor: 'background.paper',
                         height: 350,
-                        maxWidth: 700,
+                        width: `calc(100vw - 15vw)`,
+                        maxWidth: 650,
                     }}
                 >
                     {rail ? (


huayunh (commented):
When hovered, your drawer did a "content-cover" animation and did not shrink the page body. However, when I clicked on the hamburger menu, the drawer did a partial content push. It shrank the page body width, but not the app bar. You would want to make the behavior more consistent. This can potentially be fixed in only the drawer playground. But if you think there might be a bug for the persistent drawer component itself, please log an issue.

Here's a screen recording to help illustrate what I am talking about:

https://github.com/etn-ccis/blui-react-component-library/assets/8997218/5fb26937-a5b3-4183-abbd-ea2826ac27f7

JeffGreiner-eaton:
> When hovered, your drawer did a "content-cover" animation and did not shrink the page body. However, when I clicked on the hamburger menu, the drawer did a partial content push. It shrank the page body width, but not the app bar. You would want to make the behavior more consistent. This can potentially be fixed in only the drawer playground. But if you think there might be a bug for the persistent drawer component itself, please log an issue.
> 
> Here's a screen recording to help illustrate what I am talking about:
> 
>  Untitled.mov

Yes I see the same on the deployed master branch dev docs as well. https://brightlayer-ui-components.github.io/react/components/drawer/playground
I will create issue

huayunh (approved):

PR #855, opened 2024-02-14
Remove divider from snippets & update content

JeffGreiner-eaton:
Fixes https://github.com/etn-ccis/blui-react-component-library/issues/775.

- remove divider from DrawerNavGroup
- update code content for copy
-

-

#### To Test:

-

-


surajeaton (commented):
I think the issue logged was about the wrong prop name. It should be `titleDivider` instead of divider.

We may need to replace that and not remove.

surajeaton (approved):
Looks good.

